PR #4, opened 2019-11-03
Change module path to include full source path

jen20:
As per other Terraform providers, the module path should include `github.com/terraform-providers` as a prefix.


ldhenry:
Thanks for your help, @jen20!

Issue #23, opened 2020-04-21
launchdarkly plagin exited with exit status 2


Ri0n:
Hi
during terraform plan it starts handling the flags (75 flags + 4 per environement configuration with fallthrough to default variation setting for all them), but somewhere in the middle I get this:

```
Error: rpc error: code = Canceled desc = context canceled



Error: rpc error: code = Unavailable desc = transport is closing


Releasing state lock. This may take a few moments...
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4: 
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4: goroutine 189 [running]:
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4: github.com/terraform-providers/terraform-provider-launchdarkly/launchdarkly.targetsToResourceData(0xc00060a300, 0x1, 0x4, 0xeb5ce0, 0xc0005bcf60)
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4:  /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-launchdarkly/launchdarkly/target_helper.go:62 +0x1ef
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4: github.com/terraform-providers/terraform-provider-launchdarkly/launchdarkly.resourceFeatureFlagEnvironmentRead(0xc00031b500, 0xe98b40, 0xc0001384c0, 0xc00031b500, 0x0)
2020-04-21T18:31:00.592+0300 [DEBUG] plugin.terraform-provider-launchdarkly_v1.2.1_x4:  /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-launchdarkly/launchdarkly/resource_launchdarkly_feature_flag_environment.go:171 +0x761
[...]

PR #24, opened 2020-04-23
prepare 1.2.2 release

ldhenry:
FEATURES:

- Added `/examples`, a directory containing a variety of detailed usage examples.

BUG_FIXES:

- Fix non-empty plan after creating a `launchdarkly_team_member` with a custom role.
- Handle missing user target variations in `launchdarkly_feature_flag_environment` [#23](https://github.com/terraform-providers/terraform-provider-launchdarkly/issues/23)


launchdarkly/config.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	AgentVersion = "1.2.1"
+	AgentVersion = "1.2.2"
```
sloloris, line 12: i don't know what this does but this is on purpose right?

ldhenry: Yup that sets the User Agent for all API calls. We should always try to keep this synced up with the version.

PR #52, opened 2021-03-10
Fix: rollout weights documentation.

thorfour:
The rollout weights documentation listed the required sum to be
1,000,000 but it's actual 100,000


ldhenry:
Good eye. Thanks for fixing this.

PR #55, opened 2021-04-27
Add team_members data source

patrickmarabeas:
Adds team_members data source which allows the lookup of a list of team
members by their email and returns a list of member objects. The data
source also includes an optional field "ignore_missing" which allows
for the graceful filtering of members that return an error. This will
help in the management of members who are added through SSO facilities
and not the team_member Terraform resource.

Adds:
- launchdarkly_team_members data source


ldhenry (changes requested):
Hi  @patrickmarabeas,

Thanks for submitting this, we are always grateful of outside contributions. I had one suggestion that should be addressed before releasing this data source. Let me know if you don't have time to make the change so I can get someone from our side to take over.

Thanks,
Henry

monsagri:
We've added a `team_members` resource in #77  and released it as part of [2.4.0](https://github.com/launchdarkly/terraform-provider-launchdarkly/releases/tag/v2.4.0).

I know it's been a while since the initial request, but please let us know if that works for you.

PR #56, opened 2021-04-27
Create SSO team member role resource

hsy3418:
- Create a new resource called "sso_team_member" which will be used to apply a role to an existing team member that get access from SSO, as the current team member resource can only create new user while doing role assignment. By having this resource, it can benefit to role assignment automation for the case that users are provisioned from SSO.


ldhenry (commented):
Hi @hsy3418,

Thanks for putting this up, we always appreciate external contributions. As I understand it, you are looking to update the roles of team members that were created via SSO, is that correct? 

I think this should be theoretically possible with the current provider version by utilizing a command like `terraform import launchdarkly_team_member.example 5f05565b48be0b441fb63020`, but I could see how this would be quite cumbersome. Have you tried this approach?

In any case I do think we should move forward with something like this, however what do you think about adding a attribute boolean attribute `SSO_UPDATE` to the existing team member resource instead?

Thanks,
Henry

Issue #84, opened 2022-01-28
Error with `team_member` resource: expected role to be one of [reader writer admin], got owner


kiddom-kq:
I am trying get all our users managed via TF as SSO is $$$. See LD on the wall of shame, here: https://sso.tax/.

I have the account owner declared as a resource:

```hcl
# pointyhair boss <pointyhair@some.tld>
resource "launchdarkly_team_member" "pointyhair_boss" {
  email      = "pointyhair@some.tld"
  first_name = "pointyhair"
  last_name  = "boss"
  # This generates an error
  role = "owner"

  # This does not... but it also incorrect. API docs not super clear on what (attempting) to demote would do.
  #role = "admin"
  # Note: this does not work as the error is happening in the validation phase... well before TF does a refresh
  lifecycle {
    ignore_changes = [
      role
    ]
  }
}
```

When running `tf plan` I get this:

```
│ Error: expected role to be one of [reader writer admin], got owner
```

I believe the issue is here:

https://github.com/launchdarkly/terraform-provider-launchdarkly/blob/master/launchdarkly/resource_launchdarkly_team_member.go#L49

kiddom-kq:
And just for giggles, I ran the `tf apply` anyways to see what would happen. Got this back:

```
│ Error: failed to update team member with id "6....6": 400 Bad Request: {"message":"You cannot change the account owner's role"}
```

ldhenry:
Hey @kiddom-kq,

Sorry to hear this is causing issues for you. Unfortunately there are currently several protections in place that prevent us from adding the `owner` role to the Terraform provider directly, such as:
- There can only be one owner at a time
- You cannot change the owner via API (only via session auth)
- You cannot change your own role

I'll escalate 

Have you tried your workaround with`admin`, and `ignore_changes`?:
``` hcl
# pointyhair boss <pointyhair@some.tld>
resource "launchdarkly_team_member" "pointyhair_boss" {
  email      = "pointyhair@some.tld"
  first_name = "pointyhair"
  last_name  = "boss"
  # This generates an error
  role = "admin"
  lifecycle {
    ignore_changes = [
      role
    ]
  }
}
```

Thanks,
Henry

PR #85, opened 2022-01-28
Support `team_member` role type 'owner'

kiddom-kq:
See #84


ldhenry:
Unfortunately, this will not work as the `owner` role can only be changed inside our app using session auth. I'll raise this issue to the team to see if we can find a more satisfactory solution for managing account owners via Terraform, but a change will likely need to happen inside of LaunchDarkly for this to happen.

PR #90, opened 2022-04-07
prepare 2.5.1 release

monsagri:


ldhenry (changes requested):
Hey I think this should actually v2.6.0 since it technically includes includes a new feature.

PR #99, opened 2022-05-11
(bug-fix) Update modules to accept new API header

lucywyman:
Hashicorp [recently added a Release
API](https://www.hashicorp.com/blog/announcing-the-hashicorp-releases-api)
that includes a new `Content-Type` header
`application/vnd+hashicorp.releases-api.v0+json`. This gets used
somewhere in making requests, but the `hc-install` Go library was
previously at version 0.3.1 which raised an error when it encountered
this header. V0.3.2 accepts the header, so this updates the Go modules
in this repo using `go get -u && go mod tidy`.


lucywyman:
Honestly kind of unclear to me if this will just get magically fixed in CI and I don't need to do this?

ctawiah (approved):
LGTM

lucywyman:
Opened a PR against the private repo https://github.com/hashicorp/terraform-plugin-sdk/commit/f74b5a63b9939416d8a0ad6d264285ce8d3a4fc8

PR #105, opened 2022-07-27
Fix merge issue in Circle config

ldhenry:
Somehow we ended up with the `CHECKPOINT_TIMEOUT` twice in our CIrcle config.


sloloris (approved):
lgtm

PR #111, opened 2022-08-09
Fix typo in multiple docs headers

lachlancooper:
This fixes links to these anchors from earlier in each page.


ldhenry (approved):
Thanks!

Issue #102, opened 2022-07-06
launchdarkly_feature_flag_environment: Support setting no value_type when providing empty values for clauses


nc-wittj:
In our project, we sometimes like to use the `launchdarkly_feature_flag_environment` resource to set up rules with clauses containing no values, so they are ready to fill in once we want to rollout the feature flag. 

Previously, we were on `v1.5.0`, and providing empty `values` worked fine. Now, on `v2.7.0`, I'm noticing that each time I run `terraform apply`, it tries to set `value_type = "string"` for clauses with no values. My guess is on the API clears out the `value_type` when there are no values, triggering the Terraform Provider to try to set it again next time via the defaulting logic. 

Would it be possible to add a way to set `value_type` to some `null` value when our values are empty, so we aren't caught in this cycle? For now, I think my work around will be to conditionally not create the rule when all the `values` are empty. But that doesn't tell the story as nicely when viewed in the UI.

lucywyman:
Hm...we use the `value_type` to determine whether we need to convert the string value that's provided into a boolean or number before sending it to the API, so setting it to `null` (or `'any'`) would be effectively the same as setting it to the default `string`. We could in theory not explicitly set the default in the TF plugin schema so that`value_type` is omitted from the TF state file if it's not provided, and then under the hood use `string` as a default. But someone could have the opposite problem where they aren't sure why their boolean value is being treated as a string and it'd be hard to debug...

I guess I'm wondering why it's a problem that `value_type=string` is set? It's only used internally to do the checking, and isn't included in API calls and shouldn't (as far as I can see at least) affect the actual feature flag that gets created. Is it just confusing that it's there?

nc-wittj:
> I guess I'm wondering why it's a problem that `value_type=string` is set?

The issue on our end is that _every_ time we run a `terraform apply`, it ends up trying to set the `value_type` _again_. It doesn't matter how many times we apply. And when Terraform updates `value_type`, it triggers a notification in Slack. For us, that means over 30 notifications each time we apply, creating unnecessary noise.

If you added some logic so it didn't try to keep setting the `value_type` if `values` is empty, I think that'd fix the issue for us. I'm fine with `value_type` being set; it's just an issue when we provide an empty `values` array.

Here's a simple project that reproduces the issue:
```terraform
terraform {
  required_version = "~> 1.2"

  required_providers {
    launchdarkly = {
      source = "launchdarkly/launchdarkly"
    }
  }
}

[...]

PR #113, opened 2022-08-24
prepare v2.9.1 release

ldhenry:
## [2.9.1] (August 24, 2022)

BUG FIXES:

- Fixes a bug in the `launchdarkly_feature_flag_environment` that prevented users from updating targeting rule clauses when the targeting rule was being used as the fallthrough variation with a percentage rollout.

- Fixes a bug in the `launchdarkly_feature_flag_environment` that resulted in the default `string` rule clause value type not being respected. [#102](https://github.com/launchdarkly/terraform-provider-launchdarkly/issues/102)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,56 +41,56 @@ jobs:
           command: TESTARGS="-v" make test
       - run:
           name: Test Data Sources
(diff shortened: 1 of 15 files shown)


sloloris (approved):
everything lgtm

Issue #170, opened 2023-08-02
Launch Darkly Crossplane Provider


felixlut:
Hey!

I don't know if this is the correct forum to raise this question, but here we go.

My organization is currently looking into Launch Darkly and how to roll it out to our developers. We’ve investigated this Terraform provider, and think that it fills most of our use-case, but we also been interested in the [Crossplane](https://www.crossplane.io/) project lately. It has a lot of similarities with Terraform, with the difference that you provider your resources as Kubernetes yaml files as opposed to Terraform HCL. 

One option that we are exploring is to use [upjet](https://github.com/upbound/upjet) to generate a Crossplane provider for Launch Darkly, but before doing so we wanted to reach out to you to see if you have any plans of maintaining one yourselves? 

The [upjet documentation](https://github.com/upbound/upjet/blob/main/docs/generating-a-provider.md) have a quite extensive documentation for how this process would work. 

Thanks in advance!

ldhenry:
Hey @felixlut,

I've just added an internal feature request for a Crossplane provider. As of today, we do not have any immediate plans to build a Crossplane provider but I will leave this issue open so others can upvote it.

In the meantime, have you considered using the [provider-terraform](https://github.com/upbound/provider-terraform)? I have never tried this, but it seems like it could work.

Also, if you do end up generating a LaunchDarkly provider with upjet, please share your results here. I'm curious to see how it turns out.

Thanks,
Henry

felixlut:
@ldhenry the Terraform provider is an alternative that we are considering as well. Currently we're leaning towards either that, or to try and generate our own native Crossplane provider.

Our use-case for this is to have my users (i.e., the rest of development at my org) write their feature flags as Kubernetes yaml:s and handle them in a gitops fashion, together with their regular k8s deployments. 

The issue with the terraform approach is that it still requires handling of the statefile, which we want to abstract away if possible. Terraform by nature is also quite rigid, and has a tendency to get pissy about drift, eg someone deleted something manually, and terraform notices it and crashes, as opposed to just recreating it. It's a great tool, but if you want flexibility you need to bend over backwards sometimes

ldhenry:
That makes a ton of sense. Thanks for sharing - I've added this info to the feature request.

felixlut:
For anyone interested how the experimentation went we figured out the following:

[...]

PR #178, opened 2023-08-25
update page title to adhere to what crossplane expects

felixlut:
Update the title to adhere to what [Upjet](https://github.com/upbound/upjet) expects. 

Related to #170


ldhenry (changes requested):
Hey @felixlut, our docs (including `page_title`s) were recently updated so that they are generated from the resource schema using Hashicorp's [terraform-plugin-docs](https://github.com/hashicorp/terraform-plugin-docs).  This is an approach [recommended by Hashicorp](https://developer.hashicorp.com/terraform/registry/providers/docs#generating-documentation) and taken by our peers, such as [Datadog](https://github.com/DataDog/terraform-provider-datadog/blob/master/docs/resources/api_key.md?plain=1#L3) and [JFrog](https://github.com/jfrog/terraform-provider-artifactory/blob/813325777e3835524a90ed6e56006a8cfee9810b/docs/resources/anonymous_user.md?plain=1#L3), and is likely something that more and more providers will switch to in the future. For that reason, I don't recommend making this change. Likely Upjet will need to make an adjustment to how they consume these docs.

felixlut:
@ldhenry Aight, thanks for the quick feedback. I'll check how it looks on that side instead!

felixlut:
Turns out that the latest release of upjet 4 days ago resolves this issue.

PR #180, opened 2023-08-30
Fix typo in webhook example code

lachlancooper:
This removes a duplicate `resources =` from the second statement in the launchdarkly_webhook example code.

It also formats the HCL for this example.


ldhenry (approved):
Thanks for your submission. This looks good to me!

PR #225, opened 2024-12-04
chore: set up goreleaser

sloloris:
This should set up the necessary release.yml and .goreleaser.yml files to push to the Terraform registry once we've tagged a new version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: Release
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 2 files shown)


ldhenry (changes requested):
We will also need to update the `.goreleaser.yml` file to be similar to the example one [here](https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.goreleaser.yml).

I think we will also need to add a `terraform-registry-manifest.json` file as well.

ldhenry (approved):
This looks good to me though you'll need to pull main again and update the PR title.

PR #226, opened 2024-12-04
chore: set up release-please

sloloris:
This action should leverage [Google's release-please](https://github.com/googleapis/release-please-action) to scan conventional commits made since the last release, auto-update the changelog, and creating / updating a release PR as you continue to merge new commits. 

When you merge the release PR, it should create a tag based on the version auto-incremented according to the conventional commits it contains that are prefixed with the following:
- `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
- `feat:` which represents a new feature, and correlates to a SemVer minor.
- `feat!:`, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the !) and will result in a SemVer major.

If i've understood correctly, other conventional commits will not be included in the changelog (example `chore:`).

The tag that release-please creates when you merge a release PR will be picked up by goreleaser for release.


.github/workflows/tag-for-release.yml
```diff
@@ -0,0 +1,19 @@
...
+  contents: write
+  pull-requests: write
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
```
ldhenry, line 16: Please [pin this to the latest SHA instead of using a version tag](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash).

ldhenry (changes requested):

PR #229, opened 2024-12-04
chore: add test workflow

sloloris:
**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees
For more information on how to build, test, and release, see the [internal README on Confluence](https://launchdarkly.atlassian.net/wiki/spaces/ENG/pages/457379492).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,7 @@
 # Terraform Provider testing workflow.
-name: Tests
+name: "Tests"
 
 on:
-  pull_request:
-    paths-ignore:
-      - 'README.md'
   push:
     paths-ignore:
       - 'README.md'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@
 name: "Tests"
 
 on:
-  push:
+  pull_request:
     paths-ignore:
       - 'README.md'
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,6 @@ name: Test
 
 on:
   pull_request:
-    branches: [main]
   push:
     branches: [main]
 
@@ -27,8 +26,6 @@ jobs:
       - run: go build -v .
       - name: Run linters
         uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
-        with:
-          version: v6.1.1
 
   generate:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 # Terraform Provider testing workflow.
-name: "Tests"
+name: Tests
 
 on:
   pull_request:
@@ -52,6 +52,7 @@ jobs:
 
   # Run acceptance tests in a matrix with Terraform CLI versions
   test:
+    name: Acceptance Tests
     runs-on: ubuntu-latest
     timeout-minutes: 10
     env: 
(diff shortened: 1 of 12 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+
+  # Run acceptance tests in a matrix with Terraform CLI versions
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env: 
+      TF_ACC: "1"
+      LAUNCHDARKLY_ACCESS_TOKEN: ${{ secrets.LAUNCHDARKLY_ACCESS_TOKEN }}
+    strategy:
+      matrix:
```
ldhenry, line 61: I think we might be running into issue because conflicting tests are running in parallel. Doing this will force them to run in series.
```suggestion
      max-parallel: 1 # Ensures only one job runs at a time
      matrix:
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+      - run: make generate
+      - name: git diff
+        run: |
+          git diff --compact-summary --exit-code || \
+            (echo; echo "Unexpected difference in directories after code generation. Run 'make generate' command and commit."; exit 1)
+
+  # Run acceptance tests in a matrix with Terraform CLI versions
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
ldhenry, line 56: I think this might be the limit for each matrix test case so 30 mins feels a little long.
```suggestion
    timeout-minutes: 10
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
+# Terraform Provider testing workflow.
+name: "Tests"
```
ldhenry, line 2: I think it might be nice to have a more specific name since we will use this as a required status check.
```suggestion
name: "Acceptance tests"
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
```
ldhenry, line 7: We should also have this run on merges to `main`. Also, we probably need to have this run even when the readme changes because we are going to use it as a required status check for merging PRs.
```suggestion
on:
  pull_request:
  push:
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+      - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+      - run: go mod download
+      - run: go build -v .
+      - name: Run linters
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
+        with:
+          version: latest
```
ldhenry, line 30: We should pin this to a specific version so it doesn't break in the future.

.github/workflows/test.yml
ldhenry: If the tests are passing in parallel we should leave it.

.github/workflows/test.yml
```diff
@@ -0,0 +1,92 @@
...
+          - TestAccTeamMember_CreateGeneric
+          - TestAccTeamMember_UpdateGeneric
+          - TestAccTeamMember_CreateWithCustomRole
+          - TestAccTeamMember_UpdateWithCustomRole
+          - TestAccTeam_
+          - TestAccTeamRoleMapping_
+          - TestAccWebhook
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
```
ldhenry, line 85: We should pin the latest sha's instead of version tags.

ldhenry (changes requested):

.github/workflows/test.yml
sloloris: it also includes the build and generate scripts. i think i can specifically name the tests job though

ldhenry (approved):
Nice work on this!

PR #240, opened 2024-12-06
chore: send Slack notifications when acceptance tests fail on `main`

ldhenry:
This PR sets up Technique 1 of the [Slack Send Github Action](https://github.com/slackapi/slack-github-action). This will send a webhook to invoke a Slack Workflow that will send a message to our project channel.

### Example runs
CleanShot 2024-12-06 at 3  39 05

CleanShot 2024-12-06 at 3  41 40

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,7 +110,7 @@ jobs:
 
   notify-slack-on-failure:
     name: Notify Slack on Failures
-    if: failure()
+    if: failure() && github.ref == 'refs/heads/main'
     needs:
       - build
       - generate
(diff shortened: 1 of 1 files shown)


emahannaLD (approved):

.github/workflows/test.yml
```diff
@@ -107,3 +107,20 @@ jobs:
...
+    if: failure() && github.ref == 'refs/heads/main'
+    needs:
+      - build
+      - generate
+      - test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send GitHub trigger payload to Slack Workflow Builder
+        id: slack
+        uses: slackapi/slack-github-action@485a9d42d3a73031f12ec201c457e2162c45d02d # v2.0.0
```
emahannaLD, line 122: where is this id from? Just curious for my own knowledege

ldhenry: This is just the [commit SHA of v2.0.0](https://github.com/slackapi/slack-github-action/commit/485a9d42d3a73031f12ec201c457e2162c45d02d). [Our security team recommends that we use commit SHAs instead of tags because tags can be changed to introduce malicious code](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash).

PR #246, opened 2024-12-09
chore: create release-please-config.json

sloloris:
I am trying to test [this PR](https://github.com/launchdarkly/terraform-provider-launchdarkly/pull/245) but the release please action is trying to pull this configuration file from main. So I guess I have to merge this into main first.


.github/workflows/tag-for-release.yml
```diff
@@ -1,20 +1,44 @@
...
 
 jobs:
-  release-please:
+  release-package:
     runs-on: ubuntu-latest
+    permissions: 
+      contents: write
+      pull-requests: write
+    outputs:
+      release-created: ${{ steps.release.outputs.release_created }}
```
ldhenry, line 12:
```suggestion
      release-created: ${{ steps.release.outputs.releases_created }}
```

ldhenry (changes requested):

.github/workflows/tag-for-release.yml
```diff
@@ -1,20 +1,44 @@
...
+      upload-tag-name: ${{ steps.release.outputs.tag_name }}
```
ldhenry, line 13: These outputs are only needed if we had a separate job that needed to reference them. Based on how you have it now, we could just get rid of them.

Alternatively, we could make the goreleaser stuff a separate job that only runs if a release is created. That might actually be a cleaner implementation because we wouldn't have to call `if: ${{ steps.release.outputs.releases_created == 'true' }}` in every step.

PR #272, opened 2025-01-30
fix: add random characters to name that keeps conflicting

sloloris:
This test keeps failing on our Github validation tests due to bad cleanup from previous runs causing a name conflict on the custom role. This adds some random characters to the name of the afflicted role to prevent future conflicts.


ldhenry (approved):
Looks reasonable to me.

PR #280, opened 2025-01-30
chore: fix 409s in tests

sloloris:
There are a few spots I missed previously where we were still hitting 409 conflicts when running tests because we gave custom roles the same name / key. Hopefully this is the last of them.


ldhenry (approved):
Looks good to me.

ldhenry:
Hmm, it seems like `make generate` is trying to make some changes?

PR #286, opened 2025-02-06
feat: add role attributes to custom roles

sloloris:
Updates tests and documentation.


launchdarkly/resource_launchdarkly_custom_role_test.go
```diff
@@ -145,35 +163,20 @@ func TestAccCustomRole_CreateWithStatements(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
-		},
-	})
-}
-
-func TestAccCustomRole_CreateWithNotStatements(t *testing.T) {
```
sloloris, line 152: i consolidated all the custom roles tests. since the update tests all had a create step using the same resource configs anyways, it was redundant to have separate create and update tests.

ldhenry (approved):
This looks good. Thanks for updating the docs too. Hopefully the `$${}` syntax doesn't trip too many people up.

PR #289, opened 2025-02-06
feat: add role attributes to `launchdarkly_team_member`

sloloris:
This PR defines the `role_attributes` schema block and adds a new `role_attributes` field to the `launchdarkly_team_member` resource and data source. It also adds tests and updates the docs accordingly.


launchdarkly/role_attributes_helper.go
```diff
@@ -0,0 +1,107 @@
...
+	"fmt"
+	"slices"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	ldapi "github.com/launchdarkly/api-client-go/v17"
+)
+
+func roleAttributesSchema(isDataSource bool) *schema.Schema {
+	return &schema.Schema{
+		Type: schema.TypeSet,
```
ldhenry, line 13: I think we've run into issues with the TypeSet in the past. Do you remember that too? IIRC we went with TypeList for these kind of things instead.
```suggestion
		Type: schema.TypeList,
```

sloloris: hmm i don't fully remember, no. i had issues with the type list because it's a map on the LD end so the order was irregular, which caused unexpected diffs. i could pull something more complicated like we have to enforce ordering of nested environments on projects, but this seemed easier and that also seems to periodically fail. will ping you on slack now to discuss pros/cons

launchdarkly/role_attributes_helper.go
```diff
@@ -0,0 +1,107 @@
...
+				}
+			}
+		} else {
+			if old != nil {
+				for k := range *old {
+					patch = append(patch, patchRemove(fmt.Sprintf("/roleAttributes/%s", k)))
+				}
+			}
+		}
+	}
```
ldhenry, line 105: This feels a little more complicated than it needs to be. Did you try just sending something like:
```
{
  "action": "replace",
  "path": "/roleAttributes",
  "value": <The entire new role attributes map>
}
```

ldhenry (changes requested):

launchdarkly/role_attributes_helper.go
ldhenry: This seems to work for me. Can you give it a try:
```
func getRoleAttributePatches(d *schema.ResourceData) []ldapi.PatchOperation {
	var patch []ldapi.PatchOperation
	if o, n := d.GetChange(ROLE_ATTRIBUTES); o != n {
		new := roleAttributesFromResourceData(d.Get(ROLE_ATTRIBUTES).(*schema.Set).List())
		if new != nil {
			patch = append(patch, patchReplace("/roleAttributes", new))
		} else {
			patch = append(patch, patchReplace("/roleAttributes", make(map[string][]string)))
		}
	}
	return patch
}
```

ldhenry: Looks like the tests are passing [here](https://github.com/launchdarkly/terraform-provider-launchdarkly/pull/291). Feel free to merge it into this PR if it helps.

ldhenry (changes requested):

launchdarkly/role_attributes_helper.go
sloloris: great, done

ldhenry (approved):
Looks good to me.

Issue #223, opened 2024-11-21
Receiving `approvalSettings and resourceApprovalSettings service properties must match` error on resource creation without any changes in environment provisioning


fcma-creid1:
Provider version: 2.17.0

We use the LaunchDarkly Terraform Provider for creating our projects and adding environment overrides when needed. If environment overrides aren't specified, we use the expected defaults.

We haven't changed anything around our defaults, but now getting an odd error from the provider when adding a new project:

```hcl
module.launchdarkly_projects.launchdarkly_project.my-project["my-new-api"]: Creating...
╷
│ Error: failed to update environment with key "production" for project: "my-new-api": 400 Bad Request: {"message":"approvalSettings and resourceApprovalSettings service properties must match"}
│
│   with module.launchdarkly_projects.launchdarkly_project.my-project["my-new-api"],
│   on ..\..\src\project.tf line 1, in resource "launchdarkly_project" "my-project":
```

I can add the project manually, and set the ServiceNow approvals the way I would expect on the Production environment. I can't find any references to the error I'm getting, and as best as I can tell, the format of the block is what I'd expect - pulled from the debug logs running terraform apply, I see this production environment block for the tainted resource:

```json
              {
                "api_key": "",
                "approval_settings": [
                  {
                    "can_apply_declined_changes": true,
                    "can_review_own_request": false,
                    "min_num_approvals": 1,
                    "required": true,
                    "required_approval_tags": null,
                    "service_config": {
                      "detail_column": "justification",
                      "template": "2d0d5f001bafb010308ec9d2604bcb80"
                    },
                    "service_kind": "servicenow"
                  }
                ],
                "client_side_id": "",
                "color": "417505",
                "confirm_changes": true,
                "default_track_events": false,
                "default_ttl": 0,
                "key": "production",
                "mobile_key": "",
                "name": "Production",
                "require_comments": true,
                "secure_mode": false,
                "tags": null
              }
```

The ServiceNow block looks like I'd expect, and I'm also able to manually configure the project in the LD UI. Unfortunately, Terraform taints it and wants to tear it down and recreate it (to failure) every time.

Appreciate any insights or context that would help!

fcma-ahill1:
Bumping this! 

I'm working on tackling this issue. From what I can tell, it appears there are underlying API changes that aren't reflected in the provider. 

This issue is also present in the latest version of the provider (2.21.2).

This is unfortunately preventing us from using the provider at all :(

[...]

PR #295, opened 2025-02-13
feat: add auto apply to env approvals

sloloris:
Add the auto-apply feature to environment approvals. This can be used when configuring approval integrations such as ServiceNow.


ldhenry:
It looks like we're going to need to hook up ServiceNow in our acceptance test account. I can do this now.

ldhenry:
> It looks like we're going to need to hook up ServiceNow in our acceptance test account. I can do this now.

This is done now. I re-ran the tests but it's still failing because a template is not being specificed. Also, you are likely to run into the error reported in #223 until the fix in [REL-5885](https://launchdarkly.atlassian.net/browse/REL-5885) is merged.

sloloris:
> It looks like we're going to need to hook up ServiceNow in our acceptance test account. I can do this now.

thank you!!! this is fixed, but we're now hitting the previously identified bug with the segment approvals config - i think possibly it may have been fixed for post creates but overlooked for patches. i am putting this back into draft mode while we address that first.

blaqbern:
> > It looks like we're going to need to hook up ServiceNow in our acceptance test account. I can do this now.
> 
> thank you!!! this is fixed, but we're now hitting the [previously identified bug](https://github.com/launchdarkly/terraform-provider-launchdarkly/issues/223) with the segment approvals config - i think possibly it may have been fixed for post creates but overlooked for patches. i am putting this back into draft mode while we address that first.

Hmmm. The fix I did was for PATCH environment. Sounds like maybe it didn't work. I can dig into this

ldhenry:
> > > It looks like we're going to need to hook up ServiceNow in our acceptance test account. I can do this now.
> > 
> > 
> > thank you!!! this is fixed, but we're now hitting the [previously identified bug](https://github.com/launchdarkly/terraform-provider-launchdarkly/issues/223) with the segment approvals config - i think possibly it may have been fixed for post creates but overlooked for patches. i am putting this back into draft mode while we address that first.
> 
> Hmmm. The fix I did was for PATCH environment. Sounds like maybe it didn't work. I can dig into this

@blaqbern, I just re-ran the tests and the error is:
```
{"message":"The `serviceKind` and `serviceConfig` fields for `approvalSettings` and `resourceApprovalSettings` must match"}
```
This is slightly different than what it said previously:
```
{"message":"approvalSettings and resourceApprovalSettings service properties must match"}
```

I think that means we need to make a similar change for the `serviceConfig`

ldhenry (approved):
Nice to see the tests are passing 🙌🏼

PR #296, opened 2025-02-15
fix: set `critical` property on environment resource

sloloris:
This PR fixes a bug brought to our attention by a customer via the Support team where the `critical` property was only updating as expected on environments nested on the project resource, but was not functioning as expected on the environment resources themselves.


ldhenry (approved):
Nice find. Thanks for fixing this.

PR #297, opened 2025-02-17
chore: disable release-please PR draft mode and add footer

ldhenry:
This PR disables the Release Please draft PR setting and adds a note to the PR footer explaining how to trigger the tests.

Previously we thought taking the release please PR out of draft mode would trigger our tests. Unfortunately that's not the case. The next best thing.

CleanShot 2025-02-17 at 9  58 54@2x

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -9,7 +9,8 @@
       "draft": true
     }
   },
-  "draft-pull-request": true,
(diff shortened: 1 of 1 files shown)


monsagri (approved):
Shame about the manual step, but at least it's very straightforward

PR #317, opened 2025-05-26
fix: update error messages to return properly

sloloris:
**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees


monsagri (approved):
There seem to be a lot of failing tests, but I'll approve to unblock

One question - intentionally not using `handleLdapiErr` anymore?

sloloris:
> There seem to be a lot of failing tests, but I'll approve to unblock
yeah there's something going on with the aws credential GHA but i tested locally

> One question - intentionally not using `handleLdapiErr` anymore?

we're still using `handleLdapiErr` but this was an erroneous use of it, so it was returning like `"error creating environment: %s (<nil>)"`. we actually use it in formatting the error we put on the `updateDiags` that we are then passing again to format it, but the `updateDiags` have different fields on them and so need to be formatted for return differently.

does that make sense? i feel like i wrote a lot of words but i'm not sure they fully make sense

PR #338, opened 2025-07-16
add semaphore to client

sloloris:
**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees

### DIFF
--- a/launchdarkly/config.go
+++ b/launchdarkly/config.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"golang.org/x/sync/semaphore"
+
 	retryablehttp "github.com/hashicorp/go-retryablehttp"
 	ldapi "github.com/launchdarkly/api-client-go/v17"
 )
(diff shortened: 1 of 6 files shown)


launchdarkly/config.go
```diff
@@ -18,10 +20,11 @@ import (
...
 const (
-	APIVersion     = "20240415"
-	MAX_RETRIES    = 12
-	RETRY_WAIT_MIN = 200 * time.Millisecond
-	RETRY_WAIT_MAX = 2000 * time.Millisecond
+	APIVersion              = "20240415"
+	MAX_RETRIES             = 12
+	RETRY_WAIT_MIN          = 200 * time.Millisecond
+	RETRY_WAIT_MAX          = 10000 * time.Millisecond
+	DEFAULT_MAX_CONCURRENCY = 10 // provider-specific parallelism
```
kevinkruger, line 27: Would it be worth setting the default to 1?

monsagri (approved):
lgtm

PR #347, opened 2025-09-02
Add archived field to launchdarkly_metrics resources

sasthana-ld:
Adds `archived` boolean field to `launchdarkly_metric` resource to allow users to mark metrics as archived through Terraform.

added TO-DO comments for api client since `ldapi.MetricPost` does not support `Archived` field yet so when LD api client is merged I will uncomment the archived field code in create/update operations

made changes as we have discussed in Tech spec: https://docs.google.com/document/d/1HKRRgdSnHoXjrLnzRY9_7hQw4Q5pW9Q_AQdpO0wTm9w/edit?tab=t.0#heading=h.lcgpt7k0oisq

**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees

### DIFF
--- a/docs/data-sources/metric.md
+++ b/docs/data-sources/metric.md
@@ -37,6 +37,7 @@ data "launchdarkly_metric" "example" {
 ### Read-Only
 
 - `analysis_type` (String) The method for analyzing metric events. Available choices are `mean` and `percentile`.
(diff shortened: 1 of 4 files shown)


ldhenry (commented):
Looks good so far 👍🏼

PR #354, opened 2025-10-06
feat: [RO-1835] Add release policy order resource

lucywyman:
In order (ha) to support reordering release policies from Terraform we have a 'release_policy_order' resource that accepts an ordered list of keys of scoped release policies. The provider should error if the list of keys is incomplete or includes the default key. There isn't any meaning to "deleting" and "creating" a release policy order, so both only edit the terraform statefile and don't actually take any action.

Lil demo video: https://www.loom.com/share/3de56956d2fb48769136bc656cf65e20?sid=690f6caa-9954-4b46-8959-7fbe03752fb6

---
Related Jira issue: [RO-1835: add support for POST /release-policies/order (reordering endpoint) in terraform provider](https://launchdarkly.atlassian.net/browse/RO-1835)


ember-stevens (approved):

launchdarkly/data_source_launchdarkly_release_policy_order.go
```diff
@@ -0,0 +1,55 @@
...
+			PROJECT_KEY: {
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "The project key.",
+				ValidateDiagFunc: validateKeyAndLength(1, 140),
+			},
+			RELEASE_POLICY_KEYS: {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "An ordered list of release policy keys that defines the order of release policies within the project.",
```
drewinglis, line 30:
```suggestion
				Description: "An ordered list of release policy keys that defines the priority of release policies within the project.",
```

examples/resources/launchdarkly_release_policy_order/resource.tf
```diff
@@ -0,0 +1,92 @@
...
+}
+
+resource "launchdarkly_release_policy" "policy_c" {
+  project_key    = launchdarkly_project.example.key
+  key            = "policy-c"
+  name           = "Policy C"
+  release_method = "guarded-release"
+
+  scope {
+    environment_keys = ["development"]
```
drewinglis, line 68: We should define environment this on the project as well.

launchdarkly/resource_launchdarkly_release_policy_order.go
```diff
@@ -0,0 +1,153 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      addForceNewDescription("The project key.", true),
+				ForceNew:         true,
+				ValidateDiagFunc: validateKeyAndLength(1, 140),
+			},
+			RELEASE_POLICY_KEYS: {
+				Type:        schema.TypeList,
+				Required:    true,
+				Description: "An ordered list of release policy keys that defines the order of release policies within the project.",
```
drewinglis, line 41:
```suggestion
				Description: "An ordered list of release policy keys that defines the priority of release policies within the project.",
```

launchdarkly/resource_launchdarkly_release_policy_order.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func resourceReleasePolicyOrderImport(ctx context.Context, d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	importID := d.Id()
+	if importID == "" {
+		return nil, fmt.Errorf("import ID cannot be empty")
+	}
+
+	// Parse the import ID to extract the project key
+	projectKey := strings.TrimSuffix(importID, "/release-policy-order")
```
drewinglis, line 142: Do we need the suffix here? I wouldn't expect us to need to since this resource is uniquely identified by project key.

launchdarkly/resource_launchdarkly_release_policy_order.go
```diff
@@ -0,0 +1,153 @@
...
+
+	_ = d.Set(PROJECT_KEY, projectKey)
+	d.SetId(releasePolicyOrderID(projectKey))
+
+	return []*schema.ResourceData{d}, nil
+}
+
+// releasePolicyOrderID constructs the ID for a release policy order resource
+func releasePolicyOrderID(projectKey string) string {
+	return fmt.Sprintf("%s/release-policy-order", projectKey)
```
drewinglis, line 152: Ditto re suffix.

drewinglis (commented):
I think this looks good but I'd like to understand the resource ID question before approving. I made a few wording suggestions to try to be more consistent with our public docs.

launchdarkly/resource_launchdarkly_release_policy_order.go
lucywyman: 😅 TIL! I'll update it! Thanks Drew

drewinglis (commented):
Looks like the test failure is real:
```
    data_source_launchdarkly_release_policy_order_test.go:100: Step 1/1 error: Check failed: Check 2/6 error: data.launchdarkly_release_policy_order.test: list or set attribute 'release_policy_keys' must be checked by element count key (release_policy_keys.#) or element value keys (e.g. release_policy_keys.0). Set element value checks should use TestCheckTypeSet functions instead.
```

drewinglis (approved):

PR #355, opened 2025-10-06
feat: [RO-2042] Add release policy datasource

lucywyman:
Pretty straightforward - uses the 'read' function implemented for the resource, similar to other datasources in the repo.

**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees

---
Related Jira issue: [RO-2042: Add release policies datasource in terraform](https://launchdarkly.atlassian.net/browse/RO-2042)


docs/data-sources/release_policy.md
```diff
@@ -0,0 +1,88 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "launchdarkly_release_policy Data Source - launchdarkly"
+subcategory: ""
+description: |-
+  Provides a LaunchDarkly release policy data source.
+  This data source allows you to retrieve release policy information from your LaunchDarkly project.
```
drewinglis, line 7: Is this generated from the description on line 14 in the Go code? This doesn't match -- does it need to be regenerated?

examples/data-sources/launchdarkly_release_policy/data-source.tf
```diff
@@ -0,0 +1,35 @@
...
+    release_method = data.launchdarkly_release_policy.example.release_method
+    environments   = data.launchdarkly_release_policy.example.scope[0].environment_keys
+  }
+}
+
+# Example: Check if policy is guarded release
+locals {
+  is_guarded_release = data.launchdarkly_release_policy.example.release_method == "guarded-release"
+  has_rollback = length(data.launchdarkly_release_policy.example.guarded_release_config) > 0 ? 
+    data.launchdarkly_release_policy.example.guarded_release_config[0].rollback_on_regression : false
```
drewinglis, line 20: I think this will fix the failing test.

```suggestion
  has_rollback = length(data.launchdarkly_release_policy.example.guarded_release_config) > 0
    ? data.launchdarkly_release_policy.example.guarded_release_config[0].rollback_on_regression : false
```

drewinglis (approved):
I think we should add a test for a progressive release policy as well, wdyt?

lucywyman:
> I think we should add a test for a progressive release policy as well, wdyt?

Ha! That's funny cursor added a test for that and I removed it cause I didn't think it was necessary :D but yeah I can add one!

PR #373, opened 2026-01-06
fix: improve custom_properties hashing to resolve false / missing diffs

PavelSusloparov:
Issue:
Terraform does not detect changes during the custom_properties updates.

Solution:
Use field hashing and consistent hashing.

What changed?

1. Complete field hashing: The hash now includes all fields (key, name, value), not just the key. This ensures Terraform can properly identify when any field changes.
2. Consistent hashing: By using a struct and including all fields, the hash remains consistent even when Terraform's internal TypeSet mechanism calls the function multiple times.
3. Sorted values: Values are now consistently sorted in both directions (write and read), ensuring state consistency.

**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees

### DIFF
--- a/launchdarkly/custom_properties_helper.go
+++ b/launchdarkly/custom_properties_helper.go
@@ -88,6 +88,10 @@ func customPropertiesToResourceData(customProperties map[string]ldapi.CustomProp
 		for _, v := range cp.Value {
 			values = append(values, v)
 		}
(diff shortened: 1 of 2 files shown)


sloloris (approved):
@PavelSusloparov looks great, thank you for this!!!

PR #374, opened 2026-01-14
fix: custom_properties hashing

tknopp-ld:
1. Complete field hashing: The hash now includes all fields (key, name, value), not just the key. This ensures Terraform can properly identify when any field changes.
2. Consistent hashing: By using a struct and including all fields, the hash remains consistent even when Terraform's internal TypeSet mechanism calls the function multiple times.
3. Sorted values: Values are now consistently sorted in both directions (write and read), ensuring state consistency.

**Wait!**

- Have you updated the docs?

**Testing**

## LaunchDarkly Employees

### DIFF
--- a/launchdarkly/custom_properties_helper.go
+++ b/launchdarkly/custom_properties_helper.go
@@ -88,6 +88,10 @@ func customPropertiesToResourceData(customProperties map[string]ldapi.CustomProp
 		for _, v := range cp.Value {
 			values = append(values, v)
 		}
+		// Sort the values to ensure consistency with how they're sent to the API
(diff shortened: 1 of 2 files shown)


tknopp-ld:
I made this pr to run our acceptance test suite, they don't run for forked external merge requests at the moment.

@PavelSusloparov our expert will be back in office early next week and can review your change and hopefully we can get this merged, pending review and successful tests.

sloloris (approved):
looks good to me!!! should hopefully resolve some false diffs

edit: it let me merge the original PR so i'm not sure if someone fixed the var scoping while i wasn't looking?

PR #375, opened 2026-01-20
fix: [REL-11737] Add pagination to teams resource nested fields roles and maintainers

monsagri:
This PR fixes an issue caused by us not paginating through the list of custom roles assigned to a team. This meant that if a team had more than 25 custom roles assigned, any additional custom roles would not be read by TF and added to state, leading to misleading plans and TF state.

We're changing the way we read custom roles for teams by using the dedicated GetTeamRoles endpoint instead of just expanding roles on the get teams endpoint.

**NB: Implements the same fix for team maintainers count - was previously also using expanded field which only returns up to 25 maintainers**

**NB: Also updates our Access Token valid version to include and use the most recent version `20240415`** 

Also adds a number of unit tests to cover new/changed behaviour.

---
Related Jira issue: [REL-11737: Configuration drift and state synchronization issues with Terraform provider](https://launchdarkly.atlassian.net/browse/REL-11737)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,7 @@ website/build
 website/node_modules
 
 .vscode/
+
+local-testing/
+
+.DS_Store
--- a/launchdarkly/data_source_launchdarkly_team.go
+++ b/launchdarkly/data_source_launchdarkly_team.go
@@ -106,13 +106,22 @@ func dataSourceTeamRead(ctx context.Context, d *schema.ResourceData, meta interf
 
 	// Fetch all custom role keys with pagination
 	// The expand=roles parameter only returns the first page (default 25 items)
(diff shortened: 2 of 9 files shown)


sheenacarswell (approved):
This all looks sound to me, with good tests.

Given that I haven't reviewed anything in this repo before, so I lack context, you may want to want for approval from @sloloris too

sloloris (commented):
looks great! super nitpicky suggestion that we still get the team itself in the mapping resource, just to ensure it still exists before we call the roles. i realize that the roles call probably already returns 404 or the like if the team itself does not exist, but it results in clearer error logging if we check the team first. 

also, any chance we could add a quick role pagination test or two for the team and role mapping resources?

and one other small nitpicky suggestion - the PR title will end up in the changelog, so maybe we make it more active

PR #379, opened 2026-01-27
feat: [REL-11851] update tf provider preview to support project level view requirement

monsagri:
Adds support for new Views feature - project level requirement for setting View linkage requirement at creation for Flags and Segments

---
Related Jira issue: [REL-11851: Update TF Provider preview to support project level view requirement](https://launchdarkly.atlassian.net/browse/REL-11851)


sloloris (changes requested):
looks good to me! can we just get a test with require association true and false for feature flag and segment creation? we can just set them in the project post helper iirc

ember-stevens (approved):

sloloris (approved):
lgtm thanks for tests

PR #384, opened 2026-02-03
fix: [REL-12001] incorrect header for views requests and add tests

monsagri:
We weren't properly setting the terraform provider header in our custom requests for views which makes analytics harder.

---
Related Jira issue: [REL-12001: Fix Bug with incorrect userAgent header in TF Beta Provider](https://launchdarkly.atlassian.net/browse/REL-12001)


sloloris (approved):
nice catch

PR #386, opened 2026-02-10
feat: [REL-12106] add launchdarkly_view_filter_links resource

monsagri:
WIP - want to run tests remotely

Adds support for linking resources to Views using filter statements instead of just by resource key

---
Related Jira issue: [REL-12106: Support full set of Flag and Segment filters for View Resource linking](https://launchdarkly.atlassian.net/browse/REL-12106)


launchdarkly/resource_launchdarkly_view_filter_links.go
```diff
@@ -0,0 +1,248 @@
...
+			FLAG_FILTER: {
+				Type:         schema.TypeString,
+				Optional:     true,
+				AtLeastOneOf: []string{FLAG_FILTER, SEGMENT_FILTER},
+				Description:  "A filter expression to match feature flags for linking to the view. Uses the same filter syntax as the flag list API endpoint (e.g. `tags:frontend`, `status:active`).",
+			},
+			SEGMENT_FILTER: {
+				Type:         schema.TypeString,
+				Optional:     true,
+				AtLeastOneOf: []string{FLAG_FILTER, SEGMENT_FILTER},
```
sloloris: I think you might want to use `RequiredWith` here. I think `AtLeastOneOf` will enforce setting a segment filter on all filter link resources but not enforce the co-setting of the segment filter with the segment env ID

launchdarkly/resource_launchdarkly_view_filter_links.go
```diff
@@ -0,0 +1,248 @@
...
+	betaClient, err := newBetaClient(client.apiKey, client.apiHost, false, DEFAULT_HTTP_TIMEOUT_S, DEFAULT_MAX_CONCURRENCY)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	projectKey := d.Get(PROJECT_KEY).(string)
+	viewKey := d.Get(VIEW_KEY).(string)
+
+	// Validate that segment_filter_environment_id is set when segment_filter is used
+	segmentFilterEnvId := d.Get(SEGMENT_FILTER_ENVIRONMENT_ID).(string)
```
sloloris: `RequiredWith` should omit the need for this check

launchdarkly/resource_launchdarkly_view_filter_links.go
```diff
@@ -0,0 +1,248 @@
...
+
+func resourceViewFilterLinksExists(d *schema.ResourceData, metaRaw interface{}) (bool, error) {
+	client := metaRaw.(*Client)
+	betaClient, err := newBetaClient(client.apiKey, client.apiHost, false, DEFAULT_HTTP_TIMEOUT_S, DEFAULT_MAX_CONCURRENCY)
+	if err != nil {
+		return false, err
+	}
+	projectKey := d.Get(PROJECT_KEY).(string)
+	viewKey := d.Get(VIEW_KEY).(string)
+	return viewExists(projectKey, viewKey, betaClient)
```
sloloris: i feel like here we maybe also want to check that the filter string exists rather than just the view itself

launchdarkly/resource_launchdarkly_view_filter_links_test.go
```diff
@@ -0,0 +1,420 @@
...
+			},
+			{
+				Config: fmt.Sprintf(testAccViewFilterLinksUpdateFlagFilter, projectName, projectKey, maintainerId),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckViewForFilterLinksExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, FLAG_FILTER, "tags:filter-test-v2"),
+					// After update: old filter flags should be unlinked, new filter flags should be linked
+					testAccCheckViewLinksAPIState(projectKey, "test-view", []string{"filter-test-flag-3"}),
+				),
+			},
```
sloloris: you can easily add the test case of removing links by just running a third step with `testAccViewFilterLinksCreateFlagFilter` again. also i would recommend having an `ImportStateVerify` step to ensure your plan would be empty as expected if you ran it after each apply

sloloris (changes requested):

launchdarkly/resource_launchdarkly_view_filter_links_test.go
```diff
@@ -0,0 +1,420 @@
...
+				Config: fmt.Sprintf(testAccViewFilterLinksRemoveOneFilter, projectName, projectKey, maintainerId),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckViewForFilterLinksExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, FLAG_FILTER, "tags:both-filter-test"),
+					// Segments should be unlinked after removing the segment_filter
+					testAccCheckViewLinksSegmentsAPIState(projectKey, "test-view", []string{}),
+					// Flags should still be linked
+					testAccCheckViewLinksAPIState(projectKey, "test-view", []string{"both-filter-flag-1"}),
+				),
+			},
```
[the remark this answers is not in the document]
monsagri: Not sure what you mean, as in apply a view filter link based on a tag, then apply a change to remove the tag from a flag, and then apply the filter link again?

sloloris (approved):
i'm still not entirely sure about the filter strings in state thing but happy to sign off for beta

PR #389, opened 2026-02-19
feat: Add support for AI Configs and Segment approval settings in environment resource

ldhenry:
Originally created by @tarqd. We still need to decide if this is the right approach. More information on Slack [here](https://launchdarkly.slack.com/archives/CLR408M4Y/p1771443898193139).

- Added validation to `approval_settings` that produces an error when `service_kind` (with non-default value) or `service_config` are used with `resource_kind` values other than `"flag"`. These fields are only supported for flag approval settings. Previously, these fields were silently ignored for segment and aiconfig approval settings, which could lead to unexpected behavior.


monsagri:
I'm closing old/outdated PRs as we've released V3 and fully migrated to the provider plugin model - Segment Approval settings are actually covered now, AI Config ones would be a good addition

PR #406, opened 2026-03-26
fix: [RO-2946] Update generated code to match expectation

drewinglis:
Does this depend on the API key version that you use? I couldn't replicate the exact output of CI locally.

To fix this I uncommented the `--compact-summary` argument in the CI workflow, saved the output diff as a patch file, and then `git apply`'d the patch.

---
Related Jira issue: [RO-2946: Fix generated code in terraform-provider-launchdarkly](https://launchdarkly.atlassian.net/browse/RO-2946)


monsagri:
> Does this depend on the API key version that you use? I couldn't replicate the exact output of CI locally.
> 
> To fix this I uncommented the `--compact-summary` argument in the CI workflow, saved the output diff as a patch file, and then `git apply`'d the patch.
> 
> Related Jira issue: [RO-2946: Fix generated code in terraform-provider-launchdarkly](https://launchdarkly.atlassian.net/browse/RO-2946)

I don't think it depends on the API version, but it depends on which integrations are turned on for the account you are using as the API key is used to retrieve all integration manifests via the API and build integration definitions for the TF provider from that.

I'll double check what the token we use in CI is set up as, but I'd expect it to simply be a token that has *all* available integrations enabled

ember-stevens (approved):

PR #410, opened 2026-04-30
feat: add deprecated field to feature flag schema

n-smolders:
This PR adds the deprecated field to the feature_flag resource

- Added deprecated field to keys
- Added patch call for `deprecated`
- Update docs
- Added field tests

**Testing**

- Tested by making a feature flag deprecated and running `terraform apply`
- Added tests for the new field


docs/resources/feature_flag.md
```diff
@@ -139,6 +139,7 @@ resource "launchdarkly_feature_flag" "mobile_app_feature" {
 - `client_side_availability` (Block List) (see [below for nested schema](#nestedblock--client_side_availability))
 - `custom_properties` (Block Set, Max: 64) List of nested blocks describing the feature flag's [custom properties](https://docs.launchdarkly.com/home/connecting/custom-properties) (see [below for nested schema](#nestedblock--custom_properties))
 - `defaults` (Block List, Max: 1) A block containing the indices of the variations to be used as the default on and off variations in all new environments. Flag configurations in existing environments will not be changed nor updated if the configuration block is removed. (see [below for nested schema](#nestedblock--defaults))
+- `deprecated` (Boolean) Specifies whether the flag is deprecated or not. Note that you cannot create a new flag that is deprecated, but can update a flag to be deprecated.
```
ember-stevens, line 142:
```suggestion
- `deprecated` (Boolean) Specifies whether the flag is deprecated or not. You cannot create a new flag that is deprecated, but you can deprecate an existing flag.
```

ember-stevens (approved):
✅  for the docs changes

launchdarkly/resource_launchdarkly_feature_flag_test.go
```diff
@@ -1534,6 +1554,46 @@ func testAccCheckFeatureFlagExists(resourceName string) resource.TestCheckFunc {
 
 // TestAccFeatureFlag_ViewAssociationRequired tests that creating a flag without view_keys
 // fails when the project requires view association for new flags
+func TestAccFeatureFlag_Deprecated(t *testing.T) {
```
monsagri, line 1557: Nit - the comment above doesn't match this test as it's related to the old test `TestAccFeatureFlag_ViewAssociationRequired` below. 
Please just move the comment to the right test and consider adding a short comment to this test as well (optional)

launchdarkly/feature_flags_helper.go
```diff
@@ -132,6 +132,13 @@ func baseFeatureFlagSchema(options featureFlagSchemaOptions) map[string]*schema.
 			Description: "Specifies whether the flag is archived or not. Note that you cannot create a new flag that is archived, but can update a flag to be archived.",
 			Default:     false,
 		},
+		DEPRECATED: {
```
monsagri, line 135: The generated docs are currently out of sync because docs/resources/feature_flag.md was manually edited with different wording ("You cannot create a new flag that is deprecated, but you can  deprecate an existing flag"). 
Running make generate will overwrite that back to this Description string. If you prefer the doc wording, update this Description to match, then run make generate so it stays deterministic. Something like: Description: "Specifies whether the flag is deprecated or not. You cannot create a new flag that is deprecated, but you can deprecate an existing flag."

I can make that change in case running `make generate` isn't possible!

monsagri (changes requested):
Thanks for opening this! 

I've got a tiny nit on the test comments, as well as one actual change which is due to the fact that we generate our docs based on the DESCRIPTION fields

launchdarkly/resource_launchdarkly_feature_flag_test.go
n-smolders: Thank you for highlighting this. I've moved the comment to the correct test and added a short description to the new test as well

launchdarkly/feature_flags_helper.go
n-smolders: Im unable to run `make generate` since I do not have an access_token. For me its both okay, so I reverted the description in the doc to match the string in the helper map.

launchdarkly/resource_launchdarkly_feature_flag_test.go
n-smolders: Let me know if any other changes are needed or if we are good to go

monsagri: Great, thank you!

launchdarkly/feature_flags_helper.go
monsagri: Perfect! 

I'll make sure to sort out any other make generate issues

monsagri (approved):
LGTM, thanks for raising this! 

Will make sure we bundle this into a release asap

PR #412, opened 2026-05-05
fix: remove deprecated `generate_sdk_keys` field from beta views resource

sloloris:
This field has been deprecated in the API and is no longer in use.


monsagri (approved):
lgtm - not sure if the flags test failure is flaky or if something else changed

PR #415, opened 2026-05-05
fix: prevent nil-pointer panics in optional schema attributes and harden embedded-schema (Upjet) compatibility (#387)

devopsdina:
## Summary

Closes #387.

Several optional schema attributes (`policy_statements`, `custom_roles`, `tags`,
`include_in_snippet`, `default_client_side_availability`, `policy`, etc.) were
read with naked type assertions like `d.Get(KEY).(*schema.Set)` /
`.([]interface{})`. When the runtime schema is _embedded_ — most notably under
Crossplane / Upjet, which strips deprecated attributes and may pass nil values
for unset blocks — those assertions panicked or silently coerced to wrong
defaults. The affected resources include `launchdarkly_custom_role`,
`launchdarkly_access_token`, `launchdarkly_project`,
`launchdarkly_feature_flag_environment`, and `launchdarkly_segment`.

This PR replaces the unsafe call sites with nil-safe helpers, adds an
embedded-schema compatibility shim, and tightens diagnostics so a misconfigured
or stripped schema is observable rather than fatal.

## What changed

- **New nil-safe helpers** in `launchdarkly/resource_data_helpers.go`:
  `getOptionalSet`, `optionalSetList`, `getOptionalInterfaceSlice`,
  `optionalSchemaSetFromInterface`, `interfaceSliceFromAny`,
  `optionalSetListFromAny`, `stringListFromOptionalSetValue`,
  `optionalBoolFromResourceData`, `trimmedStringAttr`. List-shaped helpers
  return an empty (non-nil) slice; bool / string helpers log at WARN on type
  mismatch so a schema regression is never silently coerced.
- **ID-derived fallbacks with explicit-error variants**: `effectiveEnvKey`
  (for `feature_flag_environment` / `segment`) and
  `effectiveCustomRoleKeyOrError`. These resolve the resource id when the
  associated attribute is absent (Upjet / Crossplane external-name flow) and
  return an error for malformed ids instead of a silent empty string.
- **Embedded-schema compatibility shim** in `launchdarkly/schema_compat.go`:
  `resourceDataSetSkipMissingKey` and `resourceDiffSetNewSkipMissingKey`
  swallow only the SDK's exact "missing schema key" errors
  (`Invalid address to set: []string{"<attr>"}` and `: invalid key: <attr>`).
  All other errors propagate. Each suppression is logged at DEBUG.
- **`launchdarkly_project` update guard**: `buildProjectUpdatePatches` no
  longer appends a fallback `replace /defaultClientSideAvailability` op when
  the runtime schema has stripped both `INCLUDE_IN_SNIPPET` and
  `DEFAULT_CLIENT_SIDE_AVAILABILITY`. Previously a "tags only" update under
  Upjet would overwrite server-side CSA defaults on every apply.
- **`customizeProjectDiff`** uses safe `cty` access (`ctyObjectGetAttr`,
  `ctyValueListElements`, `ctyBoolTrue`, `rawConfigHasAnyAttr`) so a stripped
  schema or null raw config does not panic during plan.
- **`launchdarkly_custom_role`** Read normalizes `d.SetId(customRole.Key)` so
  the Terraform id is always the canonical LD key (fixes Crossplane Observe
  matching). Create / Update accept the id as a fallback when the embedded
  schema omits `key`.
- **`launchdarkly_feature_flag_environment` / `launchdarkly_segment`** call
  `effectiveEnvKey` everywhere `env_key` is needed (Create, Read, Update,
  Delete, Exists, patch path builders). Failure messages explicitly state
  that `env_key` must be the LaunchDarkly environment **key**, not its
  display name.

## Why it fixes #387 safely

- The unsafe `.(type)` patterns are now replaced exhaustively across the
  affected resources; every removed assertion has a corresponding helper that
  tolerates `nil` and wrong types.
- `schema_compat` only swallows the two specific SDK error strings emitted
  when the runtime schema is missing the attribute. Unrelated errors (type
  coercion, validation, etc.) still surface — covered by negative tests.
- The CSA-fallback guard is gated on the schema actually exposing IIS / CSA;
  full-schema users keep the previous behavior. A regression test asserts
  both directions.

## New tests

All non-acc, no `TF_ACC` / token required.

- `launchdarkly/resource_data_helpers_test.go`
  - `TestOptionalSchemaSetFromInterface`, `TestInterfaceSliceFromAny`,
    `TestOptionalSetListAndGetOptionalInterfaceSlice_unsetOptionalBlocks`,
    `TestOptionalBoolFromResourceData`,
    `TestPoliciesFromResourceData_nilPolicyNoPanic`,
    `TestEffectiveEnvKeyFromIDOrAttr`, `TestEffectiveCustomRoleKey`.
- `launchdarkly/cty_helpers_test.go`
  - `TestCtyObjectGetAttr_*`, `TestCtyValueListElements_nullOrEmpty`,
    `TestCtyBoolTrue`, `TestRawConfigHasAnyAttr`.
- `launchdarkly/schema_compat_test.go`
  - `TestIsOmittedEmbeddedSchemaAttrErr` (true positives, false positives,
    wrapped errors), `TestIsOmittedEmbeddedSchemaAttrErr_wrapped`,
    `TestResourceDataSetSkipMissingKey_propagatesUnrelatedErrors`,
    `TestResourceDataSetSkipMissingKey_suppressesMissingKey`.
- `launchdarkly/embedded_schema_compat_test.go` (resource-level regressions)
  - `TestBuildProjectUpdatePatches_embeddedSchemaOmitsCSAFallback`,
    `TestBuildProjectUpdatePatches_fullSchemaKeepsCSAFallback`,
    `TestPolicyStatementsFromResourceData_emptyDoesNotPanic`,
    `TestAccessTokenValidate_emptyOptionalBlocksNoPanic`,
    `TestEffectiveEnvKey_*`, `TestEffectiveCustomRoleKeyOrError_errorWhenEmpty`.

## Known non-goals / unchanged behavior

- No new resources, attributes, or data sources.
- No changes to the JSON Patch payloads emitted under the full
  Terraform-CLI schema; existing acceptance tests should remain green.
- No changes to docs (no user-facing surface added).
- Team / references plumbing, `launchdarkly_environment` standalone resource
  semantics, and any Crossplane / Upjet generator code remain untouched —
  they were verified to consume the same LD identifiers and require no
  provider-side change for this fix.
- `make generate` is a no-op against this branch.

---

**Wait!**

- [x] Have you added comprehensive tests?
- [x] Have you updated relevant data sources as well as resources? _(N/A — no
  data-source surface changed.)_
- [x] Have you updated the docs? _(N/A — no user-facing surface changed.)_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -93,7 +93,7 @@ require (
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
-	github.com/zclconf/go-cty v1.14.4
+	github.com/zclconf/go-cty v1.14.4 // indirect
 	golang.org/x/crypto v0.45.0 // indirect
 	golang.org/x/exp v0.0.0-20240808152545-0cdaa3abc0fa // indirect
 	golang.org/x/mod v0.29.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -93,7 +93,7 @@ require (
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
-	github.com/zclconf/go-cty v1.14.4 // indirect
+	github.com/zclconf/go-cty v1.14.4
 	golang.org/x/crypto v0.45.0 // indirect
 	golang.org/x/exp v0.0.0-20240808152545-0cdaa3abc0fa // indirect
 	golang.org/x/mod v0.29.0 // indirect
--- a/launchdarkly/ai_config_variation_helper.go
+++ b/launchdarkly/ai_config_variation_helper.go
@@ -252,7 +252,7 @@ func aiConfigVariationRead(ctx context.Context, d *schema.ResourceData, meta int
 }
 
 func messagesFromResourceData(d *schema.ResourceData) ([]ldapi.Message, error) {
-	raw := d.Get(MESSAGES).([]interface{})
+	raw := getOptionalInterfaceSlice(d, MESSAGES)
 	messages := make([]ldapi.Message, len(raw))
 	for i, v := range raw {
 		m, ok := v.(map[string]interface{})
--- a/launchdarkly/cty_helpers.go
+++ b/launchdarkly/cty_helpers.go
@@ -1,0 +1,51 @@
+package launchdarkly
+
+import (
+	"github.com/hashicorp/go-cty/cty"
+)
+
+func ctyObjectGetAttr(config cty.Value, attr string) cty.Value {
+	if config.IsNull() || !config.IsKnown() {
+		return cty.NullVal(cty.DynamicPseudoType)
+	}
+	ty := config.Type()
+	if !ty.IsObjectType() || !ty.HasAttribute(attr) {
+		return cty.NullVal(cty.DynamicPseudoType)
+	}
+	return config.GetAttr(attr)
+}
+
+func ctyValueListElements(v cty.Value) []cty.Value {
+	if v.IsNull() || !v.IsKnown() {
+		return nil
+	}
+	return v.AsValueSlice()
+}
+
+func ctyBoolTrue(v cty.Value) bool {
+	if v.IsNull() || !v.IsKnown() || v.Type() != cty.Bool {
+		return false
+	}
+	return v.True()
+}
+
+// rawConfigHasAnyAttr reports whether the raw cty config object's *type* on a *schema.ResourceData
+// (or *schema.ResourceDiff) declares at least one of the given top-level attributes. Used to
+// detect when an embedded provider (Upjet) has stripped deprecated attributes from the runtime
+// schema so callers can avoid emitting fallback patches that would overwrite server-side state.
+//
+// The check is intentionally on the value's type, not its null-ness: helper/schema returns a
+// null value of the schema's implied object type when no live config is available, but the type
+// still carries the schema's attributes.
+func rawConfigHasAnyAttr(config cty.Value, attrs ...string) bool {
+	ty := config.Type()
+	if ty == cty.NilType || !ty.IsObjectType() {
+		return false
+	}
+	for _, a := range attrs {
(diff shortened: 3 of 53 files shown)


monsagri:
bugbot run

monsagri (approved):
@devopsdina I made a few changes to ensure the provider uses your helper in all places so we don't end up with the same issue surfacing in other resource types, apart from that I'm happy to get this merged!
Thanks for taking this on

monsagri:
### Local-test pass against prod LD

Built the provider locally (`make build`), wired it into Terraform via `dev_overrides`, and exercised two scenarios CI doesn't cover (the acc tests use clean inputs only):

**A. Optional string attrs with whitespace** — `launchdarkly_custom_role` config with `name = "  trim probe ${TS}  "` and `description = "  has space  "`. Apply succeeded; the LD API stored the values **with whitespace intact** (verified via `GET /api/v2/roles/<key>`); subsequent `terraform plan` → `No changes`.

**B. Project tags-only update under full schema** — `launchdarkly_project` with `tags = ["initial-tag"]`, then re-applied with `tags = ["initial-tag", "second-tag"]`. Live API returned `defaultClientSideAvailability = {usingMobileKey: true, usingEnvironmentId: false}` (LD defaults), confirming the `buildProjectUpdatePatches` CSA fallback emits and converges; subsequent `terraform plan` → `No changes`.

Resources were destroyed cleanly at end of run.

### Bug surfaced + fix (commit `bd6b9eb`)

The earlier `trimmedStringAttr` helper applied `strings.TrimSpace` on every optional string read (`NAME`, `DESCRIPTION`, `MAINTAINER_ID`, …). Pre-PR bare casts (`d.Get(KEY).(string)`) did not trim, so this was an unintentional behavior shift. Live prod confirmed LD does not normalize whitespace server-side — meaning trimming would cause **permanent plan drift** for any user with whitespace in their config (config `"  X  "` vs state `"X"` on every plan).

Fix:
- Renamed `trimmedStringAttr` → `optionalStringAttr` and dropped the `TrimSpace`
- Inlined `strings.TrimSpace` only at the two id-parsing sites (`effectiveEnvKey`, `effectiveCustomRoleKeyOrError`) where trimming a Terraform resource id is the intended behavior
- Sed-rename across ~40 callers
- Added `TestOptionalStringAttr_doesNotTrim` regression test asserting whitespace-preserved, missing-key, and wrong-type cases

`go build`, `go vet`, helper unit tests, and live prod plan-after-apply idempotency all clean.

PR #430, opened 2026-05-15
feat: plan-time validation for prerequisite flag destroy (#372)

monsagri:
Destroying a feature flag that another flag depends on as a prerequisite previously showed a clean plan and failed mid-apply with a 409 Conflict ("Flag is still in use as a prerequisite for the following environment/flag pairs"). SDKv2 short-circuited PlanResourceChange on destroy plans before CustomizeDiff could run, so this validation was unreachable on the legacy provider.

Now that the feature_flag resource is fully on terraform-plugin-framework (Phase 5 cutover), ResourceWithModifyPlan is invoked on destroy plans and the check moves to plan time:

- FeatureFlagResource.ModifyPlan gains a destroy branch that calls the beta FeatureFlagsBetaApi.GetDependentFlags endpoint when the plan is null and prior state is not.
- If dependents exist, raise an AddAttributeError on the key path with a bulleted list of (environment, dependent_flag) pairs and a remediation hint pointing at launchdarkly_feature_flag_environment.prerequisites.
- Non-Enterprise tokens 403 the beta endpoint; degrade to a warning so the destroy still proceeds and the apply-time 409 path remains as defence-in-depth.
- New TestAccFeatureFlag_DeletePrerequisitePlanError exercises the plan-time error via a PlanOnly step with ExpectError.

Closes #372 once verified against an Enterprise account.


sheenacarswell (approved):
👍

PR #445, opened 2026-06-10
feat: add API-coverage drift report (autogen pipeline stage 1)

monsagri:
## Summary

Stage 1 of the autogen pipeline: a deterministic drift report that diffs the endpoint families (OpenAPI tags) of LaunchDarkly's public spec against the resources and data sources registered in the provider.

- `scripts/driftreport` — Go tool in the main module; enumerates registered types at runtime via the provider's own registration lists (cannot drift from code). Exit codes: 0 clean / 1 error / 2 drift.
- `scripts/driftreport/mapping.yaml` — curated coverage contract. All 59 current spec families classified: 17 covered, 19 ignored (reasons required), 1 partial, 22 triage. Drift = new spec tag not in mapping, stale mapping entry, or registered type unreferenced by any family.
- `.github/workflows/drift-report.yml` — weekly cron + manual dispatch; report in step summary + artifact; maintains a single tracking issue labeled `api-drift`.

## Notes

- Live run against the current spec exits 0 (no drift).
- GitHub cron fires on the default branch only — the schedule arms itself when v3 lands on main; until then use `workflow_dispatch`.
- Tag-level granularity; endpoint-level coverage is a possible tool v2.

## Testing

- 6 unit tests (fixture-based drift/clean/stale detection, mapping validation, spec-parsing tolerance, live provider enumeration).
- `go build ./...`, `go vet` clean.

---
**Stack:** PR 1 of 2 (stage 1 — drift report). Stage 2 (scaffold workflow) stacks on this branch.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to a new dev/CI tool, mapping data, and workflow notifications; they do not alter provider runtime behavior or customer-facing Terraform resources.
> 
> **Overview**
> Adds **autogen pipeline stage 1**: a `scripts/driftreport` CLI that loads the public OpenAPI spec, groups paths by tag, compares them to `mapping.yaml`, and cross-checks registered Terraform resources/data sources via the provider’s registration lists. Drift is **new spec tags**, **stale mapping tags**, or **provider types not referenced** in the mapping; exit code **2** signals drift.
> 
> Introduces **`mapping.yaml`** as the coverage contract (covered / partial / ignored / triage) for current API families. **`go.mod`** promotes `gopkg.in/yaml.v3` to a direct dependency for the tool.
> 
> Adds **`.github/workflows/drift-report.yml`** (`workflow_dispatch` only for now): builds the binary (to preserve exit code 2), writes a minimal public step summary, and on drift sends a **jq-built, truncated** markdown report to a **private Slack webhook**—no artifacts or detailed report in public CI output. Unit tests cover spec parsing, mapping validation, drift detection, and live type enumeration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e34cca7f85d9624f92f8951c8a8b847b3165e5fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/drift-report.yml
```diff
@@ -0,0 +1,64 @@
...
+#
+# Diffs the endpoint families of LaunchDarkly's public OpenAPI spec against the
+# resources registered in the provider, using the curated mapping at
+# scripts/driftreport/mapping.yaml. On drift, opens (or updates) a single
+# tracking issue labeled `api-drift`.
+name: API drift report
+
+on:
+  schedule:
+    - cron: "0 8 * * 1" # weekly, Monday 08:00 UTC (default branch only)
```
monsagri: This should currently only run on preview-v3, not default branch

.github/workflows/drift-report.yml
```diff
@@ -0,0 +1,64 @@
...
+      - uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
+        with:
+          name: drift-report
+          path: drift-report.md
+      - name: Open or update tracking issue
+        if: steps.drift.outputs.drift == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
```
monsagri: We shouldn't use issues for tracking this kind of work as the repo is public and this is part of our internal development workflow - is there something we can use that won't be public facing?

scripts/driftreport/mapping.yaml
```diff
@@ -0,0 +1,270 @@
...
+#   covered — family is modeled by the listed resources/data sources
+#   partial — some endpoints modeled; note says what's missing
+#   ignored — deliberately not an IaC surface; reason required
+#   triage  — acknowledged, coverage decision pending (listed in report, not drift)
+#
+# A spec tag missing from this file is reported as drift (new API family).
+version: 1
+families:
+  - tag: AI Configs
+    status: covered
```
monsagri, line 15: If a 'family' has a status of `covered`, could the script still discover/open PRs for new resources? that would be preferable

.github/workflows/drift-report.yml
monsagri: Good catch — GitHub only evaluates `schedule:` triggers on the default branch, so the cron could never have run against `preview-v3` anyway. Removed the cron entirely in 649ecad5; the workflow is `workflow_dispatch`-only for now, and we'll re-add the schedule once the v3 line becomes `main`.

.github/workflows/drift-report.yml
monsagri: Agreed — issues are gone. Drift now goes to a private Slack channel via a Workflow Builder webhook (`SLACK_DRIFT_WEBHOOK_URL` secret), and `issues: write` was dropped from the workflow permissions (649ecad5). On top of that, 9475fc2d keeps the report body off the public run entirely: no artifact upload, the step summary only states drift yes/no, and the full report is sent to Slack as an escaped+truncated payload variable.

scripts/driftreport/mapping.yaml
monsagri: Not yet — granularity in this first pass is the OpenAPI tag (family), so a `covered` family won't surface new endpoints added within it. That endpoint-level coverage (per-resource operationId mapping, with `partial` families reporting spec operations not claimed by any resource) is the next planned enhancement on the tool, starting right after this stack lands. The `partial` status already exists in the mapping schema as the hook for it; the limitation is documented as a known gap.

PR #505, opened 2026-07-20
feat!: promote v3 to main (v3.0.0 GA)

monsagri:
## ⚠️ This PR is the v3.0.0 GA promotion. Merge with care.

Promotes the **v3 line to `main`**. The commit's tree is **exactly `preview-v3`** (verified: empty diff vs `origin/preview-v3`); it sits on top of `main` so it merges as a normal PR — no force-push, no branch-protection change.

### Merge instructions
- **Squash-merge only.** That gives `main` a single promotion commit; full v3 history stays on `preview-v3` + tags, full v2 history on the `v2` branch.
- The squash body carries `Release-As: 3.0.0`, so after merge **release-please opens a `chore: release 3.0.0` PR**.
- **Version gate:** before merging *that* release PR, confirm it proposes exactly `3.0.0` (not `2.31.0` / `3.0.0-beta.8` / `4.0.0`). Then merge → tags `v3.0.0` → goreleaser publishes assets + signed SHA256SUMS → Registry promotes to latest.
- Post-merge: `gh release edit v3.0.0 --notes-file .github/release-notes/v3.0.0.md`; delete `preview-v3` (history preserved by tags).

### Why the diff is huge
This is the whole v2→v3 delta (452 files). The content was reviewed incrementally on `preview-v3` and shipped through `3.0.0-beta.7`; this PR is the mechanical promotion, not new code.

### What's in it
Plugin Framework (protocol 6), nested-attribute syntax, 10 new resources / 9 data sources, `archive_flags_on_destroy`, `migrate-tf-syntax`, automatic state upgraders. See `docs/guides/v3-release-notes.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
> Major-version GA with widespread HCL and state-shape changes; incorrect merge or release versioning would affect every Registry consumer, though the change is an intentional promotion rather than a narrow hotfix.
> 
> **Overview**
> This PR **lands the v3 line on `main`** as **3.0.0 GA** (tree matches the former `preview-v3` line). It is a **breaking** release: **Plugin Framework / protocol 6**, nested attributes instead of blocks, removed deprecated fields, **10 new resources / 9 data sources**, automatic state upgraders, and the **`migrate-tf-syntax`** CLI (also built in **GoReleaser** as separate release archives).
> 
[...]

PR #476, opened 2026-06-29
ci(driftreport): surface triage families as one-click scaffold links

monsagri:
## Human Summary

Ensure that the "auto-dispatch" flow covers partially resource-covered features in our API, not just completely untouched ones.

## What

Adds a **block 4** to the drift report's Slack notify step (`drift-report.yml`): one message per `status: triage` family, each carrying a pre-filled `gh workflow run scaffold-resource.yml -f family='<tag>'` command + a one-click ▶ Scaffold link.

## Why

The notify step gated on drift alone (`if: drift == 'true'`). But `status: triage` families — the bulk of the V3 coverage gap (Release Management, Experimentation, the beta `Other` cluster, ~13 families) — are **not drift**: they're acknowledged-but-pending. So they never surfaced in Slack and had **no trigger path** at all, even though `scaffold-resource.yml` whole-family mode already accepts any spec tag and can build them.

This closes the gap between "the drift tool knows about a parked family" and "an operator can dispatch a scaffold for it in one click" — without changing the family's `triage` status or the human scope-decision gate.

## Changes (one file, +37/−5)

- New `triage` output on the drift step (count of `triage_families`).
- Notify step now runs on **drift OR any triage family**; renamed to `Notify Slack (drift + triage backlog)`.
- Summary message gains a triage tally + a `Triage backlog (decide scope, then dispatch): …` line; title generalized to `🔎 API coverage report` (accurate when there's no drift).
- Block 4 posts one `⚠️ Triage family: <tag>` message each, wording keeps **"decide whether it belongs in Terraform BEFORE dispatching"** explicit.

## Deliberately scoped out

- **`triage_operations`** (AgentControl optimizations / restricted models / targeting): can't be one-click scaffolded — scoped mode needs a resource-name + operations split decided first. Left for a mapping promotion (`triage_operations` → `new_resource_candidates`), which then flows through the existing `scaffoldable_resources` link path.
- **Future cron**: when a weekly schedule lands, gate block 4 behind `workflow_dispatch` so the static backlog isn't re-posted weekly (noted inline; harmless today — dispatch is the only trigger).

## Validation

- `python3 yaml.safe_load` — parses.
- Ran the exact summary `jq` + block-4 `jq` against a synthetic report: correct tally, backlog line, and per-family iteration.
- Verified `gh` command quoting for a tag with spaces/parens (`-f family='Release pipelines (beta)'`) and that `jq -nc --arg` JSON-encodes the payload safely (no injection).
- No provider/Go code touched.

[...]

PR #510, opened 2026-07-22
chore: harden autogen pipeline (restore drift-report regression, turn ceilings, defect ledger, orphan sweep)

monsagri:
## What

Audit-driven hardening of the provider autogen pipeline (drift report → scaffold → verify). Seven incremental `chore:` commits; no provider code touched.

### Regression fix 
- **Restore #476/#478 in `drift-report.yml`**: the v3 promotion (#505) replaced main's copy with the stale preview-v3 version, silently reverting triage-family Slack messages (#476) and the Workflow Builder plain-text formatting fixes (#478). Restored on top of the #505 deltas (cron, no checkout pin), and implemented the previously-noted cron gating: the static triage backlog posts on **manual dispatch only**, so the weekly cron alerts on genuine drift alone.

### Safety brakes

- **`--max-turns` on all three agent invocations** (scaffold 200 / verify 150 / triage 100 — sized from the pilot's 123 turns). `timeout-minutes` bounds wall-clock only; this bounds token spend.
- **Scaffold concurrency keyed per family** (was `resource_name || family`): a scoped dispatch and a whole-family dispatch for the same family both edit the family's `mapping.yaml` entry and now serialize.
- **Corrected the over-claiming exfil comment** in `verify-reusable.yml`: dropping `Bash(curl:*)` narrows surface but is not a security boundary (`go run` / terraform registry fallback are arbitrary egress with the LD token in env). The real mitigation — a scoped LD token (custom role limited to `tf-verify-*` + the singleton kinds) — needs an LD-account + SSM change and is **not** in this PR.

### Learning loop

- Verify agent now appends `{pr, resource, defect_categories, notes}` to a **committed ledger** (`.github/agent-prompts/verify-findings.jsonl`) and repeats the categories in `verify-result.json`; the Slack notification surfaces them per run. Turns each verify run into durable signal for periodic scaffold-prompt/skill updates (previously a manual one-off, #454).
- Fixed a stale line in the scaffold prompt claiming a branch-name fallback for verification that the workflow deliberately removed — omitting `scaffold-result.json` skips verify entirely.

### Housekeeping

- **New `sweep-verify-projects.yml`**: weekly deterministic sweep (no agent) deleting `tf-verify-pr<N>` projects whose PR is closed/merged. Keeps projects on any ambiguity; `dry_run` dispatch input.
- **`mapping.yaml` header**: documented the covered-family blind spot (tag-level claims never re-drift) + the standing promotion task to op-level `partial` for high-traffic families.

## Validation

- All 4 touched/added workflow YAMLs parse.
- `go build ./scripts/driftreport` clean; all 17 driftreport tests pass (mapping.yaml still loads).
- Slack bodies use bare URLs / plain-text commands per the #478 Workflow Builder findings.

---

> [!NOTE]
> **Medium Risk**
> Changes affect production-adjacent CI (real LD applies on verify, project deletion sweep) and agent spend; mistakes could skip verification or delete test projects, but scope is workflows/prompts only with conservative keep-on-ambiguity behavior.
> 
> **Overview**
> Hardens the **drift → scaffold → verify** autogen pipeline with workflow and agent-prompt changes only (no provider code).
> 
> **Drift report (`drift-report.yml`)** restores triage-family Slack notifications and **plain-text** scaffold commands (Workflow Builder does not render mrkdwn). Weekly cron still alerts on real drift only; **manual dispatch** also posts the static triage backlog. Summary/titles updated accordingly.
> 
> **Safety:** `--max-turns` on scaffold (200), verify (150), and triage (100). Scaffold **concurrency is per `family`**, so whole-family and scoped runs for the same family cannot race on `mapping.yaml`. Scaffold prompt states **no branch-name fallback** for stage-3 when `scaffold-result.json` is missing.
> 
> **Learning loop:** Verify agents must append **`verify-findings.jsonl`**, commit it (even on clean runs), and echo **`defect_categories`** in `verify-result.json` and Slack.
> 
> **Housekeeping:** New **`sweep-verify-projects.yml`** deletes `tf-verify-pr<N>` LaunchDarkly projects when the PR is no longer open (keeps on ambiguity; optional `dry_run`). **`mapping.yaml`** header documents the **covered-family blind spot** and the standing task to promote high-traffic families to `partial`.
> 
> **Docs:** `verify-reusable.yml` clarifies that dropping `Bash(curl:*)` is not a full security boundary vs scoped LD tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c01828a8f8eaa05a49f52adc62495876a1a262b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


monsagri:
Addressed both Bugbot findings in 699bc9b:

1. **Cron summary reposts triage backlog** — valid. The summary's triage tally + "Triage backlog:" names line are now gated on `workflow_dispatch`, consistent with the per-family triage messages. Verified the jq renders correctly in both modes.
2. **Project sweep stops at first page** — valid defensive bug. Pagination now breaks on a short page (`items < limit`) instead of trusting `totalCount`; a missing field can no longer truncate the sweep to the first 100 projects.

monsagri:
Second Bugbot round addressed in c01828a:

1. **Ledger append after push** — valid, and slightly worse than reported: step 5 also read as skippable on a clean run, which would write no ledger line at all. Reordered to ledger-append → commit → push, and made the ledger line unconditional (an empty `defect_categories` line from a clean run is signal too).
2. **Wrong projects API pagination** — **rebutted**: `offset` + `limit` are documented query params on `GET /api/v2/projects` (verified against the live `openapi.json`; the endpoint also returns `_links`, but offset paging is first-class). Kept offset paging and added a `MAX_PAGES` backstop so a hypothetically ignored offset produces a warning instead of looping to the job timeout.

sloloris (approved):

PR #514, opened 2026-07-24
fix: remove archived field from launchdarkly_view resource and data source

monsagri:
## What

Removes the `archived` field from the `launchdarkly_view` **resource** and **data source** on the v2 maintenance line.

## Why

Per [REL-14370](https://launchdarkly.atlassian.net/browse/REL-14370), LaunchDarkly has removed the ability to archive views. The `archived` attribute is therefore no longer supported by the API, so it should not be exposed by the provider.

## Changes

- `resource_launchdarkly_view.go` — drop the `ARCHIVED` schema attribute and the `d.HasChange(ARCHIVED)` patch arm.
- `data_source_launchdarkly_view.go` — drop the `ARCHIVED` schema attribute.
- `view_helper.go` — drop `Archived` from the `View` struct, `viewFromAPI`, the `d.Set(ARCHIVED, ...)` call, and `patchView`.
- Acceptance-test assertions and generated docs updated.

## State compatibility

`launchdarkly_view` is a **beta** resource. `archived` was an `Optional` attribute defaulting to `false`; existing state carrying it is dropped by Terraform on the next refresh, so no state migration is required.

## Release

Conventional `fix:` → patch release via release-please (v2 maintenance branch).

## Testing

- `go build ./...`, `go vet ./launchdarkly/`, `gofmt` — clean.
- `make generate` (docs) — clean diff (only the `archived` doc line removed).
- View unit tests pass, except `TestViewRequestsIncludeUserAgentHeader`, which **fails on clean `v2` before this change** (unrelated: the test mock omits the `generateSdkKeys` property required by api-client-go/v22). Acceptance tests (`TestAccView_*`) hit a real LaunchDarkly account and were not run in this environment.

[REL-14370]: https://launchdarkly.atlassian.net/browse/REL-14370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> The global HTTP transport rewrite affects all view-related API traffic; mis-scoping could corrupt responses, though tests limit changes to view paths. Removing `archived` is a breaking schema change for configs that still set it.
> 
> **Overview**
> Removes **`archived`** from the `launchdarkly_view` resource and data source (schema, read/update/patch paths, docs, and tests) to match LaunchDarkly dropping view archival (REL-14370). Existing state with `archived` should drop on refresh without a migration.
> 
> Adds a **retryable HTTP transport shim** that injects `archived: false` into successful JSON from `/views` and `/view-associations` responses so the pinned **api-client-go** model can still deserialize until the client is regenerated.
> 
> Also refreshes generated **Jira** integration config with optional `issueCreateUrl` and `jiraMetaUrl`, and adds unit tests for the shim and injection helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b09c6c864517f74c71047dca3d47108e55d44abe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


launchdarkly/view_helper_test.go
```diff
@@ -52,19 +52,20 @@ func TestViewRequestsIncludeUserAgentHeader(t *testing.T) {
...
+			"projectKey":      projectKey,
+			"key":             viewKey,
+			"name":            "Test View",
+			"description":     "",
+			"generateSdkKeys": false,
+			"version":         1,
+			"tags":            []string{},
+			"createdAt":       0,
+			"updatedAt":       0,
+			"archived":        false,
```
sloloris: did you also mean to remove archived here?

monsagri: yes thanks, good catch!

launchdarkly/config.go
```diff
@@ -134,7 +137,84 @@ func newRetryableClient(retryPolicy retryablehttp.CheckRetry) *http.Client {
...
+	if base == nil {
+		base = http.DefaultTransport
+	}
+	resp, err := base.RoundTrip(req)
+	if err != nil || resp == nil {
+		return resp, err
+	}
+	// Only touch successful JSON view responses.
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 || !strings.Contains(req.URL.Path, "/views") {
+		return resp, nil
```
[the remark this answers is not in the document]
monsagri: Fixed in 23832443 — broadened the shim to also match `/view-associations` (GetLinkedViews), which `segment_view_keys`/`feature_flag_view_keys` reads hit. Verified by the green TestAccView + TestAccSegment acceptance jobs on this PR.

sloloris (approved):
do we have any plans to regen the client anytime soon? or is it a breaking change to remove the code reference to a defunct field? (although are we still in beta for views?)

monsagri:
> do we have any plans to regen the client anytime soon? or is it a breaking change to remove the code reference to a defunct field? (although are we still in beta for views?)

It doesn't matter in this case as you need to really dig to even find the `archived` field and in the entire time views have been available not a single customer has used the archival functionality. 


But yeah, we should get better at releasing new clients more regularly.

PR #518, opened 2026-07-27
fix: attach AgentControl variation tools via the tools field

tracisiebel:
## What

A user reported `tool_keys` on `launchdarkly_ai_config_variation` applies cleanly but tools never actually attach. **Confirmed and fixed.**

**Root cause:** the API documents two write fields — `toolKeys` ([string], "latest version of the tool will be used") and `tools` ([{key, version}]). The server **ignores `toolKeys` entirely** (proven by the API-side assertion in this PR: `API reports attached tools [], want [key]`) but honors `tools`. The provider was sending `toolKeys`.

**Fix:** resolve each `tool_keys` entry to its current version via `GetAITool` and send `tools`. No HCL changes for users — `tool_keys = [...]` now actually attaches. Removal is an explicit `tool_keys = []` (the attribute is Optional+Computed, so omitting it preserves prior state).

**Why our tests never caught it:** `TestAccAIConfigVariation_WithToolKeys` asserted Terraform state, and the read's preserve-prior fallback echoes the planned value into state when the API returns no tools — green whether or not attachment happened. This PR replaces that with `testAccCheckAIConfigVariationToolsAttached`, which asserts through the API and cannot be masked, now covering create, import, update-to-two-tools, and remove-all.

**Verified live:** the create-path fix went green on the API-side assertion (run on e019ae1); the update/removal coverage lands in this PR's latest run.

## For the API team

`toolKeys` is documented in the public OpenAPI spec on both `postAIConfigVariation` and `patchAIConfigVariation` but is a no-op server-side. Either wire it up or remove it from the spec — as-is it's a silent-failure trap for every API consumer (this provider shipped v3.0.0–v3.1.x with it).

---

> [!NOTE]
> **Medium Risk**
> Create/update now perform extra AgentControl API calls per tool key and fail if a referenced tool is missing; behavior change fixes real mis-attachment but adds dependency on tool lookup at apply time.
> 
> **Overview**
> Fixes **`tool_keys`** on **`launchdarkly_ai_config_variation`** applying in Terraform without tools actually attaching in AgentControl.
> 
> **Create/update** no longer send the API’s **`toolKeys`** field (accepted but ineffective). The provider now resolves each key with **`GetAITool`**, builds **`tools`** entries with explicit versions, and sends those on POST/PATCH. Clearing **`tool_keys`** sends an empty **`tools`** array so associations are removed.
> 
> Acceptance tests for **`tool_keys`** now cover add, expand, and remove, and assert attached tools via the API (not only state), avoiding false greens from the read path that preserves prior **`tool_keys`** when the API returns no tools.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0beab71e36dabd3a298c2bbe028e5801bf200af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/launchdarkly/resource_ai_config_variation_framework.go
+++ b/launchdarkly/resource_ai_config_variation_framework.go
@@ -312,9 +312,6 @@ func (r *AIConfigVariationResource) Create(ctx context.Context, req resource.Cre
 
 	toolKeys, d := stringSliceFromSet(ctx, plan.ToolKeys)
 	resp.Diagnostics.Append(d...)
-	if len(toolKeys) > 0 {
-		post.ToolKeys = toolKeys
-	}
 
 	judges, d := variationJudgesFromMap(ctx, plan.Judges)
 	resp.Diagnostics.Append(d...)
@@ -326,6 +323,15 @@ func (r *AIConfigVariationResource) Create(ctx context.Context, req resource.Cre
 
 	if resp.Diagnostics.HasError() {
 		return
+	}
+
+	if len(toolKeys) > 0 {
+		tools, terr := r.resolveVariationTools(projectKey, toolKeys)
+		if terr != nil {
+			addLdapiError(&resp.Diagnostics, fmt.Sprintf("failed to resolve tool versions for AgentControl config variation %q in project %q", variationKey, projectKey), terr)
+			return
+		}
+		post.Tools = tools
 	}
 
 	err := r.client.withConcurrency(r.client.ctx, func() error {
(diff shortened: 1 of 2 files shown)


tracisiebel:
**Verification complete** ✅ — full run green on `d0beab7` (all shards). Tool attachment proven through API-side assertions at every step: create (1 tool), import round-trip, update (2 tools), remove-all via `tool_keys = []`. Ready for review.

tknopp-ld (approved):

PR #522, opened 2026-08-03
fix: validate view keys before flag/segment create

monsagri:
Jira: REL-15191

Follow-up to a customer thread about implementing Views at scale. An engineer lost significant time to a mistyped view key, so the theme here is making that class of mistake cheap to diagnose and hard to make.

## `view_keys` validation moved ahead of create

A flag or segment created with `view_keys` carries those keys in the create request body. When a key did not resolve, the failure surfaced as the raw API error from the POST — no indication of which key was wrong or what to do. The provider already validated view keys, but only on the reconcile path, which create never reached.

Validation now runs before the create call, through a shared `validateViewKeysExist` helper used by both paths. The diagnostic names the offending key and points at the fix: reference the view rather than repeating its key as a string literal, or read it with a `launchdarkly_view` data source when another configuration owns it. It also notes that view keys are case-sensitive.

**This is deliberately not a plan-time check.** The provider only ever sees a single resource's plan, so it cannot distinguish a typo from a view a sibling resource is about to create in the same apply — a plan-time lookup would fail every green-field apply of a correct configuration. Users who want plan-time validation should use the `launchdarkly_view` data source, which asserts the view already exists.

## Examples and docs now reference views instead of repeating keys

Every `view_keys` / `view_key` in the examples was a string literal, so we were teaching the pattern that causes the problem. With a literal there is no dependency edge between the view and the flag, and Terraform is free to create the flag first. Updated across `launchdarkly_feature_flag`, `launchdarkly_segment`, `launchdarkly_view_links` and `launchdarkly_view_filter_links`, with the two "managed in a module" examples switched to the data source to show the cross-state pattern.

Also removed `generate_sdk_keys` from the `launchdarkly_view` example — the attribute was removed from the schema in #412 but stayed in the published docs, so copying the example produced `Unsupported argument`.

## Beta clients now honor provider configuration

Separate commit (`dbf353c8`). Every beta-API client was built with `DEFAULT_HTTP_TIMEOUT_S` / `DEFAULT_MAX_CONCURRENCY` rather than the operator's values, so `max_concurrency = 8` still serialized beta requests (the default is 1) and `http_timeout` was pinned to 20s. Affects views, environments, segments, sdk keys, release policies, metric groups, flag import configurations, big segment store integrations, integration delivery configurations and ai agent graphs.

The derived client keeps its own semaphore rather than sharing the parent's — sharing would deadlock a beta call issued from inside a `withConcurrency` closure. A saturated provider can therefore have up to 2x `max_concurrency` in flight; the 429 backoff absorbs it. Default of 1 is unchanged, so no behavior change unless `max_concurrency` was raised.

## CI coverage gap fixed

`TestAccFeatureFlagViewKeys` added to the acceptance matrix. The existing `TestAccFeatureFlag_` entry has a trailing underscore, so it never matched `TestAccFeatureFlagViewKeys_*` — those three tests, including `_NonexistentView`, have never run in CI. `TestAccSegment` and `TestAccView` do match their respective tests.

`TestAccSegmentViewKeys_NonexistentView` expected the raw `not found` API error and now expects the provider's own diagnostic. That expectation change is itself evidence of the bug being fixed.

## Testing

- `make fmt`, `go build ./...`, `go vet ./...` clean; each commit builds standalone.
- `go generate .` idempotent, so the generate-diff gate passes.
- New unit tests: `TestValidateViewKeysExist` (including a case-sensitivity case) and `TestBetaClientFromConfigInheritsProviderSettings`.
- `make test` has 6 failures, all pre-existing and reproduced on `main`: 5 require `LAUNCHDARKLY_ACCESS_TOKEN`, and `Test404RetryClient/max_retries_exceeded` races the 20s client timeout against 12 retries.
- Acceptance tests have not been run locally — relying on CI for `TestAccFeatureFlagViewKeys`, `TestAccSegment`, `TestAccView`.

## Reviewer notes

- The beta-client commit touches 17 files across non-view resources. Mechanical, but wider blast radius than the views work implies.
- AI Configs still have no `view_keys` and no project-gate guard. If `require_view_association_for_new_flags` applies to AI Configs server-side, they may be uncreatable in a gated project. Unverified, out of scope here.

---

> [!NOTE]
> **Medium Risk**
> Beta-client wiring touches many resources and can allow up to 2× `max_concurrency` in flight; view validation adds pre-create API lookups but improves failure modes without changing successful apply behavior.
> 
> **Overview**
> Makes mistyped or missing **view** keys fail with a clear provider diagnostic instead of an opaque API error when creating flags or segments with `view_keys`.
> 
> **Pre-create validation** adds shared `validateViewKeysExist`, run before the create POST and on update when `view_keys` change (skipped on create reconcile to avoid duplicate GETs). Errors name the bad key and recommend `launchdarkly_view` / `data.launchdarkly_view` references and note case sensitivity. This stays apply-time only so sibling resources creating views in the same apply still work.
> 
> **Docs and examples** replace string-literal `view_keys` / `view_key` with resource or data source references across feature flags, segments, `view_links`, and `view_filter_links`; schema text matches. Removes stale `generate_sdk_keys` from the view example.
> 
> **Provider HTTP client** introduces `Client.betaClientFromConfig()` so all beta API clients inherit the operator’s `http_timeout` and `max_concurrency` instead of hardcoded defaults; call sites across views, SDK keys, environments, and other beta resources are updated accordingly.
> 
> **CI** adds `TestAccFeatureFlagViewKeys` to the acceptance matrix (not matched by `TestAccFeatureFlag_`). Acceptance fixtures gain project environments; segment nonexistent-view test expects the new diagnostic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1dda29c05a682bf8b974087acac0c5d3462fd278. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sloloris (approved):
stamp

PR #546, opened 2026-09-02
feat: add launchdarkly_team_members batch resource (BIS-3201)

alohaninja:
## What

New resource `launchdarkly_team_members`: invites up to 50 members in a single `POST /api/v2/members` request, then assigns their teams with one grouped `PATCH /api/v2/teams/{key}` per declared team.

pr546-batch-team-members

## Why

Large enterprise onboardings managed through Terraform hit member-write rate limits with the per-member `launchdarkly_team_member` resource — 50 members is 50 sequential create calls, and 429s start appearing partway through a batch. A failed apply then has to be re-run through the customer's change-management process, stretching a single team's onboarding to days or weeks. Batching the create makes onboarding one request per batch plus one per team, which removes the rate-limit exposure entirely.

## Design decisions

- **`members` is a `MapNestedAttribute` keyed by lowercase email.** Set semantics hash the whole element, so per-entry computed `id`s break applies; the map key is the natural identity. Follows the repo's documented MapNestedAttribute conventions (inner `email` pinned at plan time; new entries' `id` marked unknown on update).
- **Team assignment is decoupled from the member POST.** Inline `teamKeys` makes the API assign teams one member at a time inside the create request; large batches cannot finish before the service's request deadline, which kills the request after members were already persisted. One grouped PATCH per team keeps every request small, and a failure is scoped to one team. Request contract (pinned by tests in `team_members_api_calls_test.go`): create = 1 POST + 1 PATCH per team, constant in batch size; refresh = 1 filtered GET.
- **`adopt_existing` (default `false`).** Conflicting emails fail the apply with guidance by default; opting in takes ownership of existing members (documented as: they will be deleted on removal/destroy). This is also the recovery path when a failed apply leaves members created but unrecorded.
- **`deletion_protection` (default `true`).** Blocks destroys and whole-batch replacements at plan time (`ModifyPlan`) and again at apply time, evaluated against prior state so it cannot be disabled and bypassed in one apply. Removing entries deletes real people; the guard makes the destructive paths deliberate two-step operations.
- **Update ordering: create → patch → delete**, so a mid-apply failure leaves people with access rather than without it. Documented consequence: swapping a full batch at exact seat capacity needs a two-apply workflow.
- Mixed ownership with `launchdarkly_team_member` / `launchdarkly_team.member_ids` is documented as unsupported; `team_keys` owns only declared associations.

## Testing

- **Unit + request-contract tests:** validators, batch diffing, conflict parsing, full-replacement detection, plus a fake-API call-count suite asserting the exact request count of every operation (including a 50-member × 35-team fan-out case and 429-retry behavior).
- **Acceptance tests (5, all passing):** `TestAccTeamMembers_CreateUpdate` (lifecycle + no-op plan + full-replacement guard), `_DeletionProtection` (destroy guard + flip-then-destroy), `_AdoptExisting` (fixture seeded via raw API, not another TF resource), `_Import` (contract asserted via `ImportStateCheck` — the batch's synthetic resource-level `id` is legitimately regenerated on import, so `ImportStateVerify` cannot apply; every member-level attribute is asserted explicitly), `_ManyTeamsPerMember` (12-team fan-out converging to a stable plan). One anchored `TestAccTeamMembers_` entry registered in both CI workflow matrices (`make matrixcheck` passes).
- **Scale validation:** 700 members across 14 batches + 700 teams applied and converged against a local backend, including a stress batch with 35 `team_keys` per member; full teardown clean.
- **Production validation** on a dedicated test account: 50-member batch with 3 teams created in ~6s (single POST ~1.8s, team PATCHes ~0.3s), zero 429s across create / no-op plan / update / two-step destroy; update produced exactly the pinned request set (grouped team PATCH, no per-member fan-out for team-only changes).
- Docs updated (`docs/resources/team_members.md`, examples, import script) via `go generate`; new attributes all covered by the tests above.

## Notes for reviewers

- Requires an Admin-level access token (Writer is denied `createMember` and team management) — documented, matching `launchdarkly_team_member`.
- `pinMapKeysToMapKey` was generalized to `pinMapKeysToAttr` (the existing helper hardcoded the `key` attribute; this resource's natural key is `email`). Old signature preserved as a wrapper.
- `github.com/google/uuid` promoted from indirect to direct in `go.mod` (synthetic batch ID).
- No new backend endpoints; entitlement and batch-size limits are enforced server-side and surfaced as API errors.

BIS-3201

---

> [!NOTE]
> **Overview**
> Adds **`launchdarkly_team_members`**, a Terraform resource that invites up to 50 account members in one `POST /api/v2/members` call and assigns teams with one grouped `PATCH` per declared team—aimed at large onboardings that hit rate limits with per-member `launchdarkly_team_member`.
> 
> The `members` map is keyed by lowercase email, with **`adopt_existing`** (default off) for conflict handling and partial-apply recovery, and **`deletion_protection`** (default on) blocking destroys and whole-batch replacements until disabled in a prior apply. Updates run create → patch → delete; import accepts comma-separated member IDs. Shared helpers add **`pinMapKeysToAttr`** (email pinning), extensive unit/API-contract tests, acceptance tests wired into CI as `TestAccTeamMembers_`, and generated docs/examples.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 852c57f1f33a27952211235b89016a9ee2f6c429. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


alohaninja:
Thanks for the thorough review — all findings actioned in 37ebddc, with one deliberate decline:

**Taken:**
- **Custom-role caching (both findings):** the resolver is now bidirectional (`customRoleResolver`, ID↔key) and shared by the read path and the write path — `patchMemberAttributes` resolves through it, so promoting N members to a shared role costs one lookup. Uncached values resolve via one helper call per direction, making a future bulk-endpoint adoption a single-site change. Pinned by `TestAPICallCount_CustomRoleLookupsAreCachedPerOperation` (10 members sharing one role = exactly 1 role GET).
- **ModifyPlan double decode:** prior state decodes once and feeds both the unknown-id marking and the full-replacement guard.
- **Import ID validation:** `parseImportMemberIDs` validates each ID against `^[a-f0-9]{24}$` up front — a pasted email or team key now fails immediately, by name, with the expected format (unit-tested).
- **`fetchMembersByID` naming:** documented the email-keyed return contract at the doc comment (every caller joins onto the email-keyed members map, so the keying is the contract; kept the name).

**Declined — partial state on team-assignment failure:** writing state alongside a Create error makes the framework mark the resource *tainted*, and a tainted batch plans **destroy-and-recreate of real member accounts** on the next apply (and with `deletion_protection = true` captured in that state, the destroy is then refused — the operator is wedged between untaint and a protection flip). The `adopt_existing` re-apply converges without deleting anyone, so convergent-but-explicit beats seamless-but-tainted for a resource whose failure mode is offboarding humans. Compromise shipped: the failure now emits a warning at the point of error telling the operator exactly how to recover, and the non-obvious taint rationale is documented at both call sites.

Full suite re-run after the changes: unit + call-count green, 5/5 acceptance passing.

alohaninja:
bugbot run

launchdarkly/team_members_helper.go
```diff
@@ -0,0 +1,917 @@
...
+		idVal, hasID := attrs[ID]
+		if !hadID && hasID {
+			if sv, isStr := idVal.(types.String); isStr && sv.IsNull() {
+				attrs[ID] = types.StringUnknown()
+				newObj, d := types.ObjectValue(objType.AttrTypes, attrs)
+				diags.Append(d...)
+				out[key] = newObj
+				changed = true
+				continue
+			}
```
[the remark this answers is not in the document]
alohaninja: Fixed in de7a2d4 — `markNewEntriesComputedUnknown` (renamed from `markNewEntriesIDUnknown`) now marks a new entry's null `role` unknown alongside `id`, so the API-defaulted role survives the consistency check. Pinned at both layers: a unit test on the plan modifier and a new CreateUpdate acceptance step that adds a custom-roles-only member on update (verified against a live backend).

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,652 @@
...
+			return fmt.Errorf("members key %q must be lowercase", email)
+		}
+		if !isPlausibleEmail(email) {
+			return fmt.Errorf("members key %q must be an email address", email)
+		}
+		if !m.Email.IsNull() && !m.Email.IsUnknown() {
+			if got := m.Email.ValueString(); strings.ToLower(got) != email {
+				return fmt.Errorf("member %q sets email %q: email must equal its map key or be omitted", email, got)
+			}
+		}
```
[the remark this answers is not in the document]
alohaninja: Fixed in de7a2d4 — the validator now requires the inner `email` to equal its map key exactly (lowercase), rather than case-insensitively. Since the API reports emails lowercased and Read writes them back that way, exact-match validation at plan time is the honest contract; the resource is unreleased so the stricter rule breaks nobody. Unit test added for the cased-mismatch rejection.

launchdarkly/team_members_helper.go
```diff
@@ -0,0 +1,994 @@
...
+	for _, id := range strings.Split(raw, ",") {
+		trimmed := strings.TrimSpace(id)
+		if trimmed == "" {
+			continue
+		}
+		if !memberIDPattern.MatchString(trimmed) {
+			return nil, fmt.Errorf(
+				"%q is not a member ID (expected a 24-character hex ID such as 5f0cd446a77cba0b4c5644a7); "+
+					"the import ID is a comma-separated list of member IDs", trimmed)
+		}
```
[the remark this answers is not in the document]
alohaninja: Fixed in de7a2d4 — `parseImportMemberIDs` now uses the provider-wide case-insensitive `idPattern` and normalizes IDs to lowercase (the casing LaunchDarkly reports) before they reach the filter string. Unit test covers uppercase acceptance + normalization.

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  -> Note: first_name and last_name are only used when a member is created. LaunchDarkly does not allow the provider to change a member's name afterwards; the member does that themselves.
+  -> Note: Updates create new members before deleting removed ones, so swapping a full batch at exactly your seat limit fails on the create. Remove members in one apply and add in the next, or add seats.
+  -> Note: Creating a batch costs one members request plus one request per declared team; team assignment is intentionally separate from the member invite so that no single request approaches the LaunchDarkly service's request deadline. Setting the provider's http_timeout to 60 is recommended for large batches: it gives headroom while staying long enough that a slow request surfaces the service's real response instead of a client-side timeout. If an apply fails part-way, LaunchDarkly may already have created the members even though Terraform recorded nothing; re-apply with adopt_existing = true to take ownership of them and finish the work.
```
ember-stevens:
```suggestion
  -> Note: Creating a batch costs one members request plus one request per declared team. Team assignment is intentionally separate from the member invite so that no single request approaches the LaunchDarkly service's request deadline. We recommend setting the provider's http_timeout to 60 for large batches. This lets a slow request surface the service's response instead of a client-side timeout. If an apply fails partway, LaunchDarkly may have already created the members even though Terraform did not record anything. Re-apply with adopt_existing = true.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "launchdarkly_team_members Resource - launchdarkly"
+subcategory: ""
+description: |-
+  Provides a LaunchDarkly team members resource for inviting up to 50 members, with their team assignments, in a single API call.
+  This resource batches member creation into one POST /api/v2/members request instead of one request per member, which avoids the member-write rate limits that per-member resources hit on large onboardings. The members map is keyed by lowercase email.
+  -> Note: You can only manage members with "admin" level access tokens: a Writer token is denied createMember and team management (403). Use an Admin token, or a custom role granting member creation and updateTeamMembers on the declared teams. To learn more, read Managing Teams https://launchdarkly.com/docs/home/account/manage-teams.
```
ember-stevens:
```suggestion
  You can only manage members with "admin" level access tokens. A Writer token is denied createMember and team management (403). Use an Admin token or a custom role granting member creation and updateTeamMembers on the declared teams. To learn more, read Managing Teams https://launchdarkly.com/docs/home/account/manage-teams.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  -> Note: Manage any given member with either this resource or launchdarkly_team_member, never both. Likewise, if you assign teams here with team_keys, do not also manage those teams' membership with launchdarkly_team.member_ids — the two fight over the same association.
```
ember-stevens:
```suggestion
  Manage any given member with either this resource or launchdarkly_team_member, never both. Likewise, if you assign teams here with team_keys, do not also manage those teams' membership with launchdarkly_team.member_ids. This results in a conflict over the same association.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  -> Note: first_name and last_name are only used when a member is created. LaunchDarkly does not allow the provider to change a member's name afterwards; the member does that themselves.
```
ember-stevens:
```suggestion
  first_name and last_name are only used when a member is created. LaunchDarkly does not allow the provider to change a member's name afterward. The member can change their name themselves.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  -> Note: Updates create new members before deleting removed ones, so swapping a full batch at exactly your seat limit fails on the create. Remove members in one apply and add in the next, or add seats.
```
ember-stevens:
```suggestion
  Updates create new members before deleting removed ones, so swapping a full batch at your seat limit fails on the create. Remove members in one apply and add in the next, or add seats.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  -> Note: Creating a batch costs one members request plus one request per declared team; team assignment is intentionally separate from the member invite so that no single request approaches the LaunchDarkly service's request deadline. Setting the provider's http_timeout to 60 is recommended for large batches: it gives headroom while staying long enough that a slow request surfaces the service's real response instead of a client-side timeout. If an apply fails part-way, LaunchDarkly may already have created the members even though Terraform recorded nothing; re-apply with adopt_existing = true to take ownership of them and finish the work.
+  -> Note: Batch size and feature availability are enforced by the LaunchDarkly API, not by this provider: an account whose plan does not allow bulk member creation receives a 403 for any batch larger than one member (nothing is created), and team_keys requires the Teams feature — team requests return 403 on plans without it. Each invited member receives an invite email plus one added-to-team notification per team they are assigned to.
```
ember-stevens:
```suggestion
  Batch size and feature availability are enforced by the LaunchDarkly API, not by this provider. An account whose plan does not allow bulk member creation receives a 403 for any batch larger than one member (no members are created), and team_keys requires the Teams feature. Team requests return a 403 on plans without it. Each invited member receives an invitation email plus one added-to-team notification for each team they are assigned to.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  Requests per operation
+  LaunchDarkly rate limits per route, so it is worth knowing what each operation costs. Inviting members is a single request no matter how many are in the batch, plus one request per declared team, which is the point of this resource. Removing members and changing their roles still cost one request each, because the API has no bulk delete and its bulk member-patch endpoint is Enterprise-only.
```
ember-stevens:
```suggestion
  LaunchDarkly rate limits per route. Inviting members is a single request no matter how many are in the batch, plus one request per declared team. Removing members and changing their roles still cost one request each, because the API has no bulk delete and its bulk member-patch endpoint is Enterprise-only.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+  ~> Warning: Removing an entry from members deletes that member from your LaunchDarkly account, and destroying this resource deletes every member in the batch. deletion_protection (enabled by default) blocks destroys and whole-batch replacements until you disable it in a separate apply.
+---
+
+# launchdarkly_team_members (Resource)
+
+Provides a LaunchDarkly team members resource for inviting up to 50 members, with their team assignments, in a single API call.
+
+This resource batches member creation into one `POST /api/v2/members` request instead of one request per member, which avoids the member-write rate limits that per-member resources hit on large onboardings. The `members` map is keyed by lowercase email.
+
+-> **Note:** You can only manage members with "admin" level access tokens: a Writer token is denied `createMember` and team management (403). Use an Admin token, or a custom role granting member creation and `updateTeamMembers` on the declared teams. To learn more, read [Managing Teams](https://launchdarkly.com/docs/home/account/manage-teams).
```
ember-stevens:
```suggestion
You can only manage members with "admin" level access tokens. A Writer token is denied `createMember` and team management (403). Use an Admin token or a custom role granting member creation and `updateTeamMembers` on the declared teams. To learn more, read [Managing Teams](https://launchdarkly.com/docs/home/account/manage-teams).
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+-> **Note:** Manage any given member with **either** this resource or [`launchdarkly_team_member`](/docs/providers/launchdarkly/r/team_member.html), never both. Likewise, if you assign teams here with `team_keys`, do not also manage those teams' membership with `launchdarkly_team.member_ids` — the two fight over the same association.
```
ember-stevens:
```suggestion
Manage any given member with **either** this resource or [`launchdarkly_team_member`](/docs/providers/launchdarkly/r/team_member.html), never both. Likewise, if you assign teams here with `team_keys`, do not also manage those teams' membership with `launchdarkly_team.member_ids`. This results in a conflict over the same association.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+-> **Note:** `first_name` and `last_name` are only used when a member is created. LaunchDarkly does not allow the provider to change a member's name afterwards; the member does that themselves.
```
ember-stevens:
```suggestion
`first_name` and `last_name` are only used when a member is created. LaunchDarkly does not allow the provider to change a member's name afterward. The member can change their name themselves.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+-> **Note:** Updates create new members before deleting removed ones, so swapping a full batch at exactly your seat limit fails on the create. Remove members in one apply and add in the next, or add seats.
```
ember-stevens:
```suggestion
Updates create new members before deleting removed ones, so swapping a full batch at your seat limit fails on the create. Remove members in one apply and add in the next, or add seats.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+-> **Note:** Creating a batch costs one members request plus one request per declared team; team assignment is intentionally separate from the member invite so that no single request approaches the LaunchDarkly service's request deadline. Setting the provider's `http_timeout` to 60 is recommended for large batches: it gives headroom while staying long enough that a slow request surfaces the service's real response instead of a client-side timeout. If an apply fails part-way, LaunchDarkly may already have created the members even though Terraform recorded nothing; re-apply with `adopt_existing = true` to take ownership of them and finish the work.
```
ember-stevens:
```suggestion
Creating a batch costs one members request plus one request per declared team. Team assignment is intentionally separate from the member invite so that no single request approaches the LaunchDarkly service's request deadline. We recommend setting the provider's http_timeout to 60 for large batches. This lets a slow request surface the service's response instead of a client-side timeout. If an apply fails partway, LaunchDarkly may have already created the members even though Terraform did not record anything. Re-apply with adopt_existing = true.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+-> **Note:** Batch size and feature availability are enforced by the LaunchDarkly API, not by this provider: an account whose plan does not allow bulk member creation receives a 403 for any batch larger than one member (nothing is created), and `team_keys` requires the Teams feature — team requests return 403 on plans without it. Each invited member receives an invite email plus one added-to-team notification per team they are assigned to.
```
ember-stevens:
```suggestion
Batch size and feature availability are enforced by the LaunchDarkly API, not by this provider. An account whose plan does not allow bulk member creation receives a 403 for any batch larger than one member (no members are created), and `team_keys` requires the Teams feature. Team requests return 403 on plans without it. Each invited member receives an invitation email plus one added-to-team notification for each team they are assigned to.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Requests per operation
+
+LaunchDarkly rate limits per route, so it is worth knowing what each operation costs. Inviting members is a single request no matter how many are in the batch, plus one request per declared team, which is the point of this resource. Removing members and changing their roles still cost one request each, because the API has no bulk delete and its bulk member-patch endpoint is Enterprise-only.
```
ember-stevens:
```suggestion
LaunchDarkly rate limits per route. Inviting members is a single request no matter how many are in the batch, plus one request per declared team. Removing members and changing their roles still cost one request each, because the API has no bulk delete and its bulk member-patch endpoint is Enterprise-only.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+# declare one launchdarkly_team_members resource per chunk, for example with a
+# module that takes a slice of the email-keyed map.
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `members` (Attributes Map) The members to invite, keyed by lowercase email address. At most 50 entries; for larger teams, split them across multiple `launchdarkly_team_members` resources. This is the authoritative set: removing an entry deletes that member. (see [below for nested schema](#nestedatt--members))
```
ember-stevens:
```suggestion
- `members` (Attributes Map) The members to invite, keyed by lowercase email address. At most 50 entries. For larger teams, split them across multiple `launchdarkly_team_members` resources. This is the authoritative set: removing an entry deletes that member. (see [below for nested schema](#nestedatt--members))
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Optional
+
+- `adopt_existing` (Boolean) Whether to take over members who already exist in your account. When `false`, the default, applying a batch that contains an existing member's email fails and tells you which emails conflict. When `true`, those members are brought under this resource's management: their roles and team assignments are reconciled to this configuration, and they are deleted when you remove them from the batch or destroy the resource.
```
ember-stevens:
```suggestion
- `adopt_existing` (Boolean) Whether to take over members who already exist in your account. When `false`, the default, applying a batch that contains an existing member's email fails and tells you which emails conflict. When `true`, those members are brought under this resource's management. Their roles and team assignments are reconciled to this configuration, and they are deleted when you remove them from the batch or destroy the resource.
```

jdmadison (approved):

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+
+<a id="nestedatt--members"></a>
+### Nested Schema for `members`
+
+Optional:
+
+- `custom_roles` (Set of String) The list of custom role keys associated with the member. Custom roles are only available to customers on an Enterprise plan. To learn more, [read about our pricing](https://launchdarkly.com/pricing/). To upgrade your plan, [contact LaunchDarkly Sales](https://launchdarkly.com/contact-sales/).
+
+-> **Note:** each member must have either a `role` or `custom_roles` argument.
+- `email` (String) The member's email address. Must equal the map key; it defaults to the map key when omitted.
```
ember-stevens:
```suggestion
- `email` (String) The member's email address. Must equal the map key. It defaults to the map key when omitted.
```

docs/resources/team_members.md
```diff
@@ -0,0 +1,165 @@
...
+- `first_name` (String) The member's given name. Used only when the member is created; afterwards only the member can change it.
+- `last_name` (String) The member's family name. Used only when the member is created; afterwards only the member can change it.
+- `role` (String) The role associated with the member. Supported roles are `reader`, `writer`, `no_access`, or `admin`. If you don't specify a role, `reader` is assigned by default.
+- `role_attributes` (Map of List of String) A map of role attributes, keyed by the role attribute key with a string array of resource keys as each value. For example, if your policy statement defines the resource `"proj/$${roleAttribute/testAttribute}"`, the key would be `testAttribute` and the values the keys of the projects you wanted to assign access to.
+- `team_keys` (Set of String) The keys of the teams to add the member to. The teams must already exist. This resource manages only the associations you declare here: teams the member belongs to for other reasons are left alone, but removing a key you previously declared removes the member from that team.
```
ember-stevens:
```suggestion
- `team_keys` (Set of String) The keys of the teams to add the member to. The teams must already exist. This resource manages only the associations you declare here. Teams the member belongs to for other reasons are left alone, but removing a key you previously declared removes the member from that team.
```

launchdarkly/team_members_helper.go
```diff
@@ -0,0 +1,1005 @@
...
+func (r *TeamMembersResource) patchMemberAttributes(
+	ctx context.Context,
+	memberID string,
+	desired teamMembersEntryModel,
+	prior teamMembersEntryModel,
+	roles *customRoleResolver,
+) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	role := desired.Role.ValueString()
```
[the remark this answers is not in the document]
alohaninja: Fixed in f72014a — `patchMemberAttributes` now falls back to the member's current role when the desired role is null/unknown/empty, rather than sending an empty string. Deliberately not hardcoded to `reader`: adopting an admin with a custom-roles-only config would have demoted them. Pinned by `TestAPICallCount_CustomRoleOnlyPatchKeepsCurrentRole`, which asserts the PATCH body carries the prior role.

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,651 @@
...
+		resp.Diagnostics.AddError("Failed to read team members", err.Error())
+		return
+	}
+	roles := newCustomRoleResolver(r.client)
+	for email, entry := range state.Members {
+		member, found := live[email]
+		if !found {
+			// Deleted outside Terraform: drop it so the next plan recreates it.
+			delete(state.Members, email)
+			continue
```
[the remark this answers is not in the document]
alohaninja: Fixed in f72014a — `fetchMembersByID` now returns a map keyed by member ID (the stable identifier) and every read-side join uses the entry's stored ID instead of email. A member whose live email drifted from the declared key stays managed (no silent drop, no re-invite of the stale address), state keeps the declared key to honor the email==key schema contract, and Read emits a warning naming the old and new email with the remediation (rename the map key + re-apply with `adopt_existing = true`). Import keys entries by the member's current lowercase email as before.

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,651 @@
...
+func (teamMembersBatchValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var cfg teamMembersResourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &cfg)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// An unknown members map (for example built from another resource's
+	// output) cannot be checked until apply.
+	if cfg.Members == nil {
+		return
```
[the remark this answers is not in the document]
alohaninja: Fixed in f72014a — the config validator now reads `members` as a `types.Map` first and defers to apply when the value is null or unknown, only decoding elements into the native map once known. Unit-tested with a raw tftypes config carrying a wholly unknown members map (`TestBatchValidatorDefersUnknownMembersMap`).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

docs/resources/team_members.md
alohaninja: Applied in f72014a — thank you! One note on mechanics: this file is generated by tfplugindocs from the resource schema descriptions, so I applied your wording at the source strings in `resource_team_members_framework.go` and regenerated, rather than committing the suggestion directly (which the next `go generate` would have reverted).

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,675 @@
...
+			// The member changed their email in LaunchDarkly. Keep managing
+			// them under the declared key rather than dropping the entry —
+			// dropping would re-invite the stale address on the next apply
+			// while the real person silently left management.
+			resp.Diagnostics.AddWarning(
+				"Team member email changed outside Terraform",
+				fmt.Sprintf(
+					"The member declared as %q now has the email %q in LaunchDarkly. Update the members map "+
+						"key to the new email and re-apply with adopt_existing = true to keep managing them "+
+						"under their current address.",
```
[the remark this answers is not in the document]
alohaninja: Fixed in 4cfe361 — Update now excludes any deletion candidate whose member ID is held by a planned entry (`excludePlannedIDs`), so the rename remediation adopts the member and skips the stale key's delete instead of deleting the person just adopted. Unit-tested for both the rename case and genuine removals.

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,675 @@
...
+					email, strings.ToLower(member.Email),
+				),
+			)
+		}
```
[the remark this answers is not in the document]
alohaninja: Fixed in 4cfe361 (same fix as the Devin thread above): deletions are filtered against IDs held by planned entries, making the map-key rename a true rename — adopt, then skip the old key's delete. The drift warning's remediation is now safe end-to-end.

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,675 @@
...
+			return
+		}
+		if state.DeletionProtection.ValueBool() {
+			resp.Diagnostics.AddError(teamMembersDestroyBlockedSummary, teamMembersDestroyBlockedDetail)
+		}
+		return
+	}
+
+	var plan teamMembersResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
```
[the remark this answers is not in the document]
alohaninja: Fixed in 4cfe361 — ModifyPlan reads `members` as `types.Map` before decoding the full model and defers when null/unknown, matching the config validator. Added `TestModifyPlanDefersUnknownMembersMap` mirroring the validator's wholly-unknown-map test as suggested.

launchdarkly/team_members_helper.go
```diff
@@ -0,0 +1,1025 @@
...
+			}
+		}
+		teamsSet, d := setFromStringSlicePreservingPlan(ctx, stillAssigned, entry.TeamKeys)
+		diags.Append(d...)
+		entry.TeamKeys = teamsSet
+	}
+
+	attrs, d := frameworkRoleAttributesValue(ctx, member.RoleAttributes)
+	diags.Append(d...)
+	entry.RoleAttributes = attrs
```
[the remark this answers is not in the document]
alohaninja: Fixed in 4cfe361 — hydration now preserves an explicitly configured `role_attributes = {}` as an empty map instead of writing null, matching the null-versus-empty handling used for `custom_roles` and `team_keys`. Covered by `TestRefreshPreservesEmptyRoleAttributes` (empty-stays-empty and omitted-stays-null).

launchdarkly/team_members_helper.go
```diff
@@ -0,0 +1,1025 @@
...
+// email: members can change their own email address, and joining API results
+// back onto the email-keyed members map by email would silently drop (and
+// later re-invite) anyone whose live email drifted from the declared key.
+func (r *TeamMembersResource) fetchMembersByID(ids []string) (map[string]ldapi.Member, error) {
+	if len(ids) == 0 {
+		return map[string]ldapi.Member{}, nil
+	}
+	idFilter := fmt.Sprintf("id:%s", strings.Join(ids, "|"))
+	expand := "roleAttributes"
+	members, err := getMembersPaginated(r.client, &idFilter, &expand, nil, teamMemberLimit, nil)
```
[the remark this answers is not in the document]
alohaninja: Acknowledged, but declining this one on empirical evidence: `customRoles` is part of the base member representation on GET /members and is returned without the expand. The acceptance suite asserts `custom_roles.# = 1` on a custom-roles-only member immediately after hydration (passing), and full-lifecycle validation on a production account converged to empty plans for members carrying custom roles — both would fail if hydration cleared the roles. The `roleAttributes` expand is the only one this endpoint needs.

launchdarkly/resource_team_members_framework.go
```diff
@@ -0,0 +1,690 @@
...
+		return
+	}
+
+	// A removed map key whose member ID is now held by another planned entry
+	// is a rename, not a removal. This happens when a member changed their
+	// email in LaunchDarkly and the operator followed the drift warning:
+	// renaming the key diffs as remove-old + create-new, adoption resolves the
+	// new key to the SAME member, and deleting the old key's ID would delete
+	// the person that was just adopted.
+	resp.Diagnostics.Append(r.deleteMembersByID(excludePlannedIDs(diff.toDeleteIDs, plan.Members))...)
```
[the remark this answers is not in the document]
alohaninja: Addressed in 852c57f, with a deliberate scope choice. The guard itself is unchanged: at plan time a batch-wide rename is indistinguishable from a genuine wipe (new entries' IDs are unknown until adoption runs), and moving the protection decision to post-adoption would mean the replacement batch gets invited *before* a genuine full replacement is refused — create-first ordering makes that strictly worse. Partial renames (the realistic email-drift case, including several members at once) never trip the guard since they are not 100% removals. What was genuinely wrong is that the drift warning promised a remediation without mentioning the whole-batch edge — the warning and the deletion_protection docs now state explicitly that renaming every key in one apply requires the same two-step as a replacement, even though adopted members are never deleted. The two-step is protection working as designed for a plan that cannot prove it is safe.

